Issue #112, opened 2016-08-25
Do payment object validation for processing only if transaction has been set/added


butenkor:
[IllegalArgumentException](https://github.com/commercetools/commercetools-payone-integration/blob/83967e1ad4e8d44fbfcefb331c85eda3472ef60e/service/src/main/java/com/commercetools/pspadapter/payone/PaymentDispatcher.java#L33-L40) should not be thrown or logged if no valid transaction is available on payment object yet. Validation should kick in only if payment object contains transaction, which needs to be processed. During checkout it is allowed to create payment object, which is not finished / consistent (i.e.: without payment method or paymentInterface set) yet as it is also allowed by CTP API.

PR #119, opened 2016-09-14
#112#114 improve logging

mht-dotsource:
@butenkor pls review


service/src/main/java/com/commercetools/pspadapter/payone/mapping/PayoneRequestFactory.java
```diff
@@ -33,4 +36,31 @@ public CaptureRequest createCaptureRequest(final PaymentWithCartLike paymentWith
...
+        try {
+            MappingUtil.mapCustomerToRequest(request, ctPayment.getCustomer());
+        } catch (final IllegalArgumentException ex) {
+            logger.debug("Could not fully map payment with ID {} {}", paymentWithCartLike.getPayment().getId(), ex.getMessage());
+        }
+
+        try {
+            MappingUtil.mapBillingAddressToRequest(request, ctCartLike.getBillingAddress());
+        } catch (final IllegalArgumentException ex) {
+            logger.error("Could not fully map payment with ID {} {}", paymentWithCartLike.getPayment().getId(), ex.getMessage());
```
andrii-kovalenko-ct: why these 3 mappings have different log levels (_debug_ for `customer` and `shippingAddress` and _error_ for `billingAddres`), when messages are absolutely the same. Looks like the function may be re-factored to use single try..catch for all three calls.

mht-dotsource: The reason for different Log-Levels is, that only billing Address (or more specific the contained values Country and Lastname) is required. The others are optinal and sould not flood the papertrail logs.
A single try/catch could prevent the execution of the other mapping methods (case: `customer == null ` and `billingAddress != null`)

PR #121, opened 2016-10-11
Health check resource

andrii-kovalenko-ct:
@butenkor @mht-dotsource @fhaertig 
please review _Health Check_ resource implementation {{IntegrationService}} errors processing re-factoring (in case of empty or malformed requests)


functionaltests/src/test/java/specs/response/HealthResponseFixture.java
```diff
@@ -0,0 +1,36 @@
...
+@RunWith(ConcordionRunner.class)
+public class HealthResponseFixture extends BasePaymentFixture {
+
+    public String getUrl() throws Exception {
+        return getHealthUrl();
+    }
+
+    public MultiValueResult handleHealthResponse() throws Exception {
+        final HttpResponse httpResponse = Request.Get(getHealthUrl())
+                .connectTimeout(200)
```
butenkor: Isn't it to low for integration test? Please check timeouts below too and use constants which are used for production env.

service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java
```diff
@@ -59,18 +65,32 @@ public void start() {
...
         Spark.post("/payone/notification", (req, res) -> {
             LOG.debug("<- Received POST from Payone: " + req.body());
-            final Notification notification = Notification.fromKeyValueString(req.body(), "\r?\n?&");
-
             try {
+                final Notification notification = Notification.fromKeyValueString(req.body(), "\r?\n?&");
                 notificationDispatcher.dispatchNotification(notification);
-            } catch (RuntimeException ex) {
+            } catch (Exception e) {
+                LOG.error("Payone notification handling error. Request body: " + req.body(), e);
```
butenkor: could you add some java comment what validations lead to 400?

service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java
```diff
@@ -59,18 +65,32 @@ public void start() {
...
             }
             res.status(200);
             return "TSOK";
         });
 
+        // This is a temporary jerry-rig for the load balancer to check connection with the service itself.
+        // For now it just returns a JSON response {"status":200}
+        // It should be expanded to a more real health-checker service, which really performs PAYONE status check.
+        // But don't forget, a load balancer may call this URL very often (like 1 per sec),
+        // so don't make this request processor heavy, or implement is as independent background service.
```
butenkor: 👍 for useful comment

service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java
```diff
@@ -59,18 +65,32 @@ public void start() {
...
+        Route healthRoute = (req, res) -> {
+            res.status(HttpStatusCode.OK_200);
+            res.type(ContentType.APPLICATION_JSON.getMimeType());
+            return ImmutableMap.of("status", HttpStatus.OK_200);
+        };
+
+        Spark.get("/health", healthRoute, SphereJsonUtils::toJsonString);
+        Spark.get("/health/", healthRoute, SphereJsonUtils::toJsonString);
```
butenkor: do we really need this too?

functionaltests/src/test/java/specs/response/HealthResponseFixture.java
andrii-kovalenko-ct: well, IMHO for this exactly resource i think it should be like this value, since it does nothing except returning text. Guess, even in future this kind of time limit should stay there to avoid _heavy_ implementations of the resource.
I'd rather specify comment in the test why to keep this value really small than increase the timeout.

service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java
andrii-kovalenko-ct: basically, not required, but during the tests i recognised i use it with final slash from time to time, so decided to add to the processing map to avoid potential confusions.
Unfortunately, Spark doesn't have default solution to process URIs with and without final slashes.

service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java
andrii-kovalenko-ct: added 9b965acb45147905fa7c8c7fe2ee2272f71c71a5

functionaltests/src/test/java/specs/response/HealthResponseFixture.java
butenkor: I am still not sure if it is good idea as you always have some network latency which you do not take into account here. By the nature of LB the node which times out will be removed from LB cluster. I would like to omit having payment service down just because of latency issues. @svenmueller is described LB behaviour correct or you have different opinion on that?

andrii-kovalenko-ct: But this is only a test! Real `/health` source implementation has default Integration Service timeout (as i know, for Spark it is 120 sec)

service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java
butenkor: I would remove as slash has different semantics (indicates a directory) and adapt tests no to use trailing slashes

functionaltests/src/test/java/specs/response/HealthResponseFixture.java
butenkor: ok. I am not fan of flaky tests also. Heroku needs some time to fire up instance

service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java
andrii-kovalenko-ct: done 051d05266f7b9975cb24fa4806a4b389e1a57dd7 and 927adfa637918c20feb03242242f9d57f5561669

functionaltests/src/test/java/specs/response/HealthResponseFixture.java
andrii-kovalenko-ct: fixed 927adfa637918c20feb03242242f9d57f5561669

PR #130, opened 2017-01-20
Cpi 125 fix route logs

andrii-kovalenko-ct:
@butenkor @ahalberkamp 

i've fixed routes logs with the next changes:
  - update `Spark` version to use the latest API
  - root request **/** now redirected to **/health**
  - raise log level of `MatcherFilter` to _warn_ to avoid logs of unmapped routes.

### DIFF
--- a/service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java
@@ -16,14 +16,11 @@ import org.apache.http.entity.ContentType;
 import org.apache.log4j.LogManager;
 import org.apache.log4j.Logger;
 import org.eclipse.jetty.http.HttpStatus;
-import spark.Redirect;
 import spark.Spark;
 
 import javax.annotation.Nonnull;
 import java.util.ConcurrentModificationException;
 import java.util.concurrent.CompletionException;
-
-import static spark.Spark.redirect;
 
 /**
  * @author fhaertig
@@ -55,8 +52,6 @@ public class IntegrationService {
         createCustomTypes();
 
         Spark.port(port());
-
-        redirect.any("/", "/health", Redirect.Status.MOVED_PERMANENTLY);
 
         // This is a temporary jerry-rig for the load balancer to check connection with the service itself.
         // For now it just returns a JSON response {"status":200}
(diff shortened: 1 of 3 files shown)


service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java
```diff
@@ -53,6 +56,19 @@ public void start() {
 
         Spark.port(port());
 
+        redirect.any("/", "/health", Redirect.Status.MOVED_PERMANENTLY);
```
butenkor: i think it is confusing to redirect base URL always to health as it should return 404 instead

service/src/main/resources/log4j.properties
```diff
@@ -2,4 +2,7 @@ log4j.rootLogger=INFO, STDOUT
 log4j.logger.deng=INFO
 log4j.appender.STDOUT=org.apache.log4j.ConsoleAppender
 log4j.appender.STDOUT.layout=org.apache.log4j.PatternLayout
-log4j.appender.STDOUT.layout.ConversionPattern=%5p [%t] (%F:%L) - %m%n
\ No newline at end of file
+log4j.appender.STDOUT.layout.ConversionPattern=%5p [%t] (%F:%L) - %m%n
+
+# skip info logs for unmapped routes
+log4j.logger.spark.http.matching.MatcherFilter=warn
```
butenkor: 👍

service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java
andrii-kovalenko-ct: Well, i just wanted to have something on base url at least for us, like for testing. Maybe not to redirect, but just return the same JSON content as from **/health** url?

butenkor: Why we can not use ```/health``` for testing too? One day you might decide base URL has to do something special. I would not mix service with explicit ```/health``` endpoint as it is not obvious and maybe be confusing for outsiders

andrii-kovalenko-ct: Ok, straightforward question:
should / return something? IMHO - should, at least "you are doing something wrong, pls, use blah-blah". 

404 - that's really confusing and looks like something wrong...

But not a big deal - we can leave the root page empty.

butenkor: Why it is confusing - that is what health/status endpoints are for? Guess what you get if you access: https://api.twitter.com or https://graph.facebook.com? You might return 400 and provide in response info which endpoints are exposed for communication or return 404

andrii-kovalenko-ct: 6a2e8ae557e14b51947102fe68008281dace6511

butenkor:
@andrii-kovalenko-ct so now it will show 404 on base URL?

andrii-kovalenko-ct:
correct, as it was before. It won't just be logged as "unmapped route"

PR #138, opened 2017-03-16
update integration tests

andrii-kovalenko-ct:
@butenkor @ahalberkamp 
The branch took more then expected from the beginning, but theses changes are important to make the integration tests more stable

  - generate Payone `psedocardpan` on demand instead of using environment variable. This avoid the issue when credit card tests fail because of expired pan.
 
 - update publish github pages plugin and settings to make publishing test results stable. Publish not only Concordion Spec HTML pages, but also gradle test console logs/outputs
  - completely re-factor notification waiting approach:
    - don't wait for failed payments
    - increase wait timeout from 8 to 15 minutes
    - make intermediate waiting terminal output to see the progress and avoid Travis tests breaking by timeout
  
  - carry out web elements names to constants in Web-verification test (like Sofortüberweisung and Visa 3ds) for easier maintaining in the future.

  - add encrypted Travis build properties to have a possibility to maintain them in the future

PR #140, opened 2017-03-23
Cpi 139 update order payment

andrii-kovalenko-ct:
@butenkor @ahalberkamp 
after you approve https://github.com/commercetools/commercetools-payone-integration/pull/138 pls take a look for the next changes:

  - Implemented update `Order#paymentState` when payment status is changed on Payone. Appropriate documentation and tests are added. See [NotificationProcessorBase.java](https://github.com/commercetools/commercetools-payone-integration/compare/CPI-127-fix-pseudocard-pan...CPI-139-update-order-payment?expand=1#diff-6befbac7289226b4a38ab3ffed9380e2)

  - re-factored CTP requests workflow: instead of direct `CtpClient` access `PaymentService` and `OrderService` are injected. Proper tests (both unit and functional) are added too.

  - for `TransactionStatus` and `NotificationAction` added `#toString()` implementations which return lowercase names for the constants, keep the same as it comes from Payone (before it was automatically uppercases)


andrii-kovalenko-ct:
closed accidentally, the review is still required

service/src/main/java/com/commercetools/pspadapter/payone/notification/NotificationProcessorBase.java
```diff
@@ -53,29 +65,81 @@ public final void processTransactionStatusNotification(final Notification notifi
...
+     *
+     * @param order          <b>nullable</b> order to update state. If <b>null</b> - the update is skipped,
+     *                       but the incident is reported.
+     * @param updatedPayment <b>non-null</b> instance of the updated payment
+     * @return a {@link CompletionStage} with <b>nullable</b> updated {@link Order} instance. If incoming order is null
+     * or new payment state is undefined, or update is not required - the returned stage is completed and contains
+     * the same instance.
+     */
+    private CompletionStage<Order> updateOrderIfExists(Order order, Payment updatedPayment) {
+        if (order != null) {
```
butenkor: should not order be Optional here?

service/src/main/java/com/commercetools/pspadapter/payone/notification/NotificationProcessorBase.java
```diff
@@ -53,29 +65,81 @@ public final void processTransactionStatusNotification(final Notification notifi
         }
 
         try {
-            client.executeBlocking(PaymentUpdateCommand.of(payment, createPaymentUpdates(payment, notification)));
+            // 1. update payment
+            CompletionStage<Order> fullCompletionStage = getPaymentService().updatePayment(payment, createPaymentUpdates(payment, notification))
+                    // 2. when payment is updated - try to find and updated respective Order.paymentState
+                    .thenComposeAsync(this::tryToUpdateOrderByPayment);
```
butenkor: this should be optional and configurable as some projects already map payment on their own

service/src/main/java/com/commercetools/pspadapter/payone/notification/NotificationProcessorBase.java
```diff
@@ -53,29 +65,81 @@ public final void processTransactionStatusNotification(final Notification notifi
...
+                                + "The order's state remains unchanged.",
+                        updatedPayment.getId(), updatedPayment.getPaymentStatus().getInterfaceCode());
+            }
+
+            // skip update for undefined or unchanged payment state
+            if (newPaymentState != null && !newPaymentState.equals(order.getPaymentState())) {
+                return getOrderService().updateOrderPaymentState(order, newPaymentState);
+            }
+        } else {
+            LOG.error("Payment [{}] is updated, but respective order is not found!", updatedPayment.getId());
```
butenkor: This might be an issue as notification "paid/capture" might arrive before user got redirected back to shop and thus before order creation. In this case order payment status will never change. Different async process based on interface interaction messages and order created messages might be more suitable here.

service/src/main/java/com/commercetools/pspadapter/payone/notification/NotificationProcessorBase.java
andrii-kovalenko-ct: Kind of holly-war topic, but at least the main developer of Java `Optional` recommends not to use it as a method argument:

> You should almost never use it as a field of something or a method parameter

http://stackoverflow.com/questions/26327957/should-java-8-getters-return-optional-type/26328555#26328555

service/src/main/java/com/commercetools/pspadapter/payone/mapping/order/DefaultPaymentToOrderStateMapper.java
```diff
@@ -0,0 +1,48 @@
...
+ *     <li><i>paid</i>, <i>capture</i> and <i>transfer</i> are mapped to {@link PaymentState#PAID}</li>
+ *     <li><i>failed</i> and <i>cancelation</i> are mapped to {@link PaymentState#FAILED}</li>
+ *     <li>the rest is mapped to <b>null</b></li>
+ * </ul>
+ */
+public class DefaultPaymentToOrderStateMapper implements PaymentToOrderStateMapper {
+
+    private static final Map<String, PaymentState> MAP;
+
+    static {
```
andrii-kovalenko-ct: an old habit, when `ImmutableMap` is not available
fixed 1d16091

service/src/main/java/com/commercetools/pspadapter/payone/ServiceFactory.java
```diff
@@ -116,19 +127,53 @@ public static void main(String [] args) throws SchedulerException, MalformedURLE
...
-    public BlockingSphereClient createCommercetoolsClient() {
-        final SphereClientFactory sphereClientFactory = SphereClientFactory.of();
-        return BlockingSphereClient.of(
-                sphereClientFactory.createClient(
-                        serviceConfig.getCtProjectKey(),
-                        serviceConfig.getCtClientId(),
-                        serviceConfig.getCtClientSecret()),
-                DEFAULT_CTP_CLIENT_TIMEOUT,
-                TimeUnit.SECONDS);
+    synchronized
```
andrii-kovalenko-ct: i reviewed it and found you are right: since we always need all of these services initialized (except optional `OrderService`, but it is lightweight anyway) it's easier to initialize them directly in the constructor and avoid lazy initialization 

fixed in 4d29a22

service/src/main/java/com/commercetools/pspadapter/payone/ServiceFactory.java
```diff
@@ -52,6 +54,15 @@
 
     private final ServiceConfig serviceConfig;
     private final PropertyProvider propertyProvider;
+    private static final String PAYONE_INTERFACE_NAME = PAYONE;
+
+    private BlockingSphereClient blockingSphereClient = null;
```
andrii-kovalenko-ct: fixed in 4d29a22

service/src/main/java/com/commercetools/pspadapter/payone/notification/NotificationProcessorBase.java
andrii-kovalenko-ct: changed logging level: 10624b0

As for "Different async process based on interface interaction messages and order created messages": guess it is not a business of this service, this service should work only with payone notifications. But looks like it is a business of a shop which creates the order: if a payment is already charged/authorized/paid - create an order with preset `paymentState`

service/src/main/java/com/commercetools/pspadapter/payone/notification/NotificationProcessorBase.java
butenkor: In this case i would not call ```updateOrderIfExists``` at all and call it rather ```updateOrder``` which is only called if order is found.

It seems that opinion of main developer is not shared at least in official article: http://www.oracle.com/technetwork/articles/java/java8-optional-2175753.html

> its purpose is to help design more-comprehensible APIs so that by just reading the signature of a method, you can tell whether you can expect an optional value. This forces you to actively unwrap an Optional to deal with the absence of a value

But lets not open holly-war. I am fine with null check also.

service/src/main/java/com/commercetools/pspadapter/payone/notification/NotificationProcessorBase.java
andrii-kovalenko-ct: added in 3ffe570
tests added in 172b658
documentation 5c592c5

andrii-kovalenko-ct:
@butenkor 
i've added fixes, pls, re-review

andrii-kovalenko-ct:
released https://github.com/commercetools/commercetools-payone-integration/releases/tag/v1.1.0

PR #142, opened 2017-03-27
Cpi 141 update jvm sdk

andrii-kovalenko-ct:
@butenkor @ahalberkamp 
just short update CT JVM SDK version from 1.3 to 1.13

PR #144, opened 2017-03-29
Cpi 133 refactor logging

andrii-kovalenko-ct:
**this PR expected to be reviewed after https://github.com/commercetools/commercetools-payone-integration/pull/140 and https://github.com/commercetools/commercetools-payone-integration/pull/142 approval**

@butenkor @ahalberkamp 

@LEQADA @heshamMassoud @lojzatran : these changes are not service specific, so might be also reviewed by you

updates:
  - avoid direct _log4j_ logging references - use _slf4j_ facade instead (see `PaymentDispatcher`,
 `ScheduledJobFactory`, `ScheduledJob`)
    - replace _log4j_ implementation and config with _logback_ one
      (see [`build.gradle`](https://github.com/commercetools/commercetools-payone-integration/pull/144/files#diff-c197962302397baf3a4cc36463dce5eaR21) and [logback.xml](https://github.com/commercetools/commercetools-payone-integration/pull/144/files#diff-7204b486ab832be7e19dae2d2af31a49))

  - setup gradle application plugin to easily run the app from IDE without manual environment variables setup and debugger attaching ([`build.gradle`](https://github.com/commercetools/commercetools-payone-integration/pull/144/files#diff-c197962302397baf3a4cc36463dce5eaR65))
  - avoid hardcoding and define the application version at runtime ([`build.gradle`](https://github.com/commercetools/commercetools-payone-integration/pull/144/files#diff-c197962302397baf3a4cc36463dce5eaR91) and [`PropertyProvider`](https://github.com/commercetools/commercetools-payone-integration/pull/144/files#diff-b5d64dd910cbb85f7c37fc04f12af0baR48))
  - re-factor the application runtime properties definition to the next order (see [`PropertyProvider`](https://github.com/commercetools/commercetools-payone-integration/pull/144/files#diff-b5d64dd910cbb85f7c37fc04f12af0baR72):
    - java command line arguments (aka `-D` arguments)
    - environment variables
   - hardcoded values from `PropertyProvider`


andrii-kovalenko-ct:
@butenkor @ahalberkamp
#140 and #142 are approved now, could you review this PR (@LEQADA @heshamMassoud @lojzatran : the PR is not specific, but common approaches are implemented, so it might be also interesting for you, see the PR description)

build.gradle
```diff
@@ -80,20 +84,80 @@ project(":service") {
...
+ *     Opposite to {@link #assertExecutableSpecRequirementsForTests(org.gradle.api.Project, org.gradle.api.tasks.testing.Test)}
+ *     this function is used for the service runtime.
+ *
+ * @param project gradle project from which to take the properties
+ * @return a list of java runtime argument in <i>-Dkey=value</i> syntax.
+ *
+ * @throws GradleException if any required system property or environment variable is undefined or empty
+ * @see #assertExecutableSpecRequirementsForTests(org.gradle.api.Project, org.gradle.api.tasks.testing.Test)
+ */
+static List<String> assertAndGetServiceProperties(Project project) {
```
butenkor: what is the use case to read arguments like ```-Dkey=value``` and not directly from env vars. Also why this validation is not executed in the service which requires them but here during build process?

functionaltests/src/test/java/util/WebDriverSofortueberweisung.java
```diff
@@ -31,7 +31,7 @@
     private final String tan;
 
     public WebDriverSofortueberweisung(final String pin, final String tan) {
-        super(BrowserVersion.FIREFOX_38, true);
+        super(BrowserVersion.FIREFOX_45, true);
```
butenkor: not FIREFOX_52?

build.gradle
andrii-kovalenko-ct: Now the app requires 7 mandatory environment variables to run the app, but IDEs (at least Intellij IDEA, but guess the other too) have *very* unhandy and manual way to manage environment variables for an application: you set them one-by-one in grid window and there is no way to automate this process ([see this SO question](http://stackoverflow.com/questions/39766723/intellij-idea-run-debug-java-application-with-a-set-of-environment-variables)]

Because of this restriction the app development has the next workflow:
  1. Build the app
  2. Export environment variables
  3. Run the app from command line with special JVM debug arguments (`java -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=1044 -jar service/build/libs/commercetools-payone-integration.jar`)
  4. Attach the debugger from IDE to the running JVM.

If you need to change one property - you edit file with exported environment variables and then repeat the steps 2-4 again.

Instead of this I'd like to make the project import-run-debug process easy to maintain (aka click-one-button), like it is set-up in the integration tests: copies values from `gradle.properties` settings to environment variables and runs the test. Unfortunately, this feature in gradle works only for test (or _exec_ task, see [this](http://stackoverflow.com/a/26634167/907576) and [this](http://stackoverflow.com/a/26634167/907576) ).

So, at last I found the gradle _application_ plugin which allows easily run a java app with some additional pre-configuration. That's why i copy the values from `gradle.properties` to java run arguments. 

Any way, i think it is _correct_ approach for normal java apps to allow override most of the properties by runtime arguments, isn't it?

functionaltests/src/test/java/util/WebDriverSofortueberweisung.java
andrii-kovalenko-ct: 45 is the latest available FF version in the latest Selenium version. I'd like to update Selenium major version to 3, but since it is a significant update - better to do this as a separate issue/PR.

build.gradle
andrii-kovalenko-ct: https://github.com/commercetools/commercetools-payone-integration/issues/153

Issue #151, opened 2017-04-12
Fix selenium test for new Sofortüberweisung web design


andrii-kovalenko-ct:
Sofortüberweisung site changed design, now the tests fail.

andrii-kovalenko-ct:
fixed https://github.com/commercetools/commercetools-payone-integration/pull/152

andrii-kovalenko-ct:
Needs changes again

ahalberkamp:
As we don`t have selenium test anymore, this can issue  be closed

PR #152, opened 2017-04-12
#151: fixed Sofortüberweisung selenium tests

andrii-kovalenko-ct:
@butenkor @ahalberkamp 
i've fixed the Sofortüberweisung tests: 

  - The problem was in wrong button search approach, so after the page design change Selenium web-driver found another button (hidden menu button) instead of _submit_ button.
  - also added failsafe check that exactly one submit button exists on the page (see `findSubmitButton()` method)

### DIFF
--- a/functionaltests/src/test/java/util/WebDriverSofortueberweisung.java
+++ b/functionaltests/src/test/java/util/WebDriverSofortueberweisung.java
@@ -12,10 +12,11 @@ import org.openqa.selenium.support.ui.ExpectedConditions;
 import org.openqa.selenium.support.ui.Wait;
 import org.openqa.selenium.support.ui.WebDriverWait;
 
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 import static org.openqa.selenium.support.ui.ExpectedConditions.not;
-import static util.constant.WebDriverCommon.CSS_SELECTOR_SUBMIT;
+import static util.constant.WebDriverCommon.XPATH_SUBMIT;
 import static util.constant.WebDriverSofortueberweisungConstants.*;
 
 /**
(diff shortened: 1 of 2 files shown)


andrii-kovalenko-ct:
@lojzatran i heard you have spare time today...
i'd appreciate if you also review it and have any concern (this is not just application specific fix)

functionaltests/src/test/java/util/constant/WebDriverCommon.java
```diff
@@ -5,7 +5,7 @@
  */
 public final class WebDriverCommon {
 
-    public static final String CSS_SELECTOR_SUBMIT = "button[type=submit]";
+    public static final String XPATH_SUBMIT = "//button[@type='submit']";
```
lojzatran: How is this different from the previous solution? If you have 2 submit buttons, it's still not identified properly, right?

functionaltests/src/test/java/util/WebDriverSofortueberweisung.java
```diff
@@ -41,8 +42,10 @@ public WebDriverSofortueberweisung(final String pin, final String tan) {
         timeouts.pageLoadTimeout(DEFAULT_TIMEOUT, TimeUnit.SECONDS);
         timeouts.setScriptTimeout(DEFAULT_TIMEOUT, TimeUnit.SECONDS);
 
+        manage().window().fullscreen();
```
lojzatran: What's the reason of this?

andrii-kovalenko-ct: In some cases it is possible that button control is hidden in a virtual browser window, and selenium will complain the button can't be clicked! 

This is not a case right now, but it protects us from future fails if they change design and the button is located on the bottom or at the right.

functionaltests/src/test/java/util/constant/WebDriverCommon.java
andrii-kovalenko-ct: correct, we would need additional logic for such a case (for instance, select by ID, name or class), but it is very unexpected right now that sofortüberweisung will have such a design.

lojzatran:
Do you have task number so I can log time?

PR #193, opened 2017-11-07
185 remove flunt hc

andrii-kovalenko-ct:
Very short PR:
improve the `HttpRequestUtil` explicitly setting connection pool size. The same sizes has `Unirest` default implementation.


service/src/main/java/com/commercetools/util/HttpRequestUtil.java
```diff
@@ -79,10 +86,19 @@
                     .setConnectTimeout(REQUEST_TIMEOUT)
                     .build())
             .setRetryHandler(HTTP_REQUEST_RETRY_ON_SOCKET_TIMEOUT)
+            .setConnectionManager(buildDefaultConnectionManager())
```
heshamMassoud: Maybe unrelated to this PR, but I don't understand this comment up there: 
`// empty implementation, we just need to use protected constructor`. Maybe make try to clarify/elaborate more?

service/src/main/java/com/commercetools/util/HttpRequestUtil.java
```diff
@@ -39,6 +40,8 @@
  * <li>retries on connections exceptions up to 3 times, if request has not been sent yet
  * (see {@link DefaultHttpRequestRetryHandler#isRequestSentRetryEnabled()}
  * and {@link #HTTP_REQUEST_RETRY_ON_SOCKET_TIMEOUT})</li>
+ * <li>connections pool is 200 connections, up to 20 per route (see {@link #CONNECTION_MAX_TOTAL}
```
heshamMassoud: Maybe makes sense to specify the reason for these numbers? 200 and 20?

service/src/main/java/com/commercetools/util/HttpRequestUtil.java
```diff
@@ -79,10 +86,19 @@
                     .setConnectTimeout(REQUEST_TIMEOUT)
                     .build())
             .setRetryHandler(HTTP_REQUEST_RETRY_ON_SOCKET_TIMEOUT)
+            .setConnectionManager(buildDefaultConnectionManager())
             .build();
 
     private static final BasicResponseHandler BASIC_RESPONSE_HANDLER = new BasicResponseHandler();
 
+    private static PoolingHttpClientConnectionManager buildDefaultConnectionManager() {
```
heshamMassoud: I would add a simple unit test to assert the default connection manager with such behaviour/values. To make sure that the that default connection manager always has such behaviour/values, so that no dev in the future changes the behaviour without knowing the side effects, if there are..

service/src/main/java/com/commercetools/util/HttpRequestUtil.java
andrii-kovalenko-ct: well, this is an inline class extension/implementation. We need to use absolutely same functionality as `DefaultHttpRequestRetryHandler`, but we can't call just `new DefaultHttpRequestRetryHandler(RETRY_TIMES, false, Arrays.asList(                    UnknownHostException.class, SSLException.class))` because this constructor is protected. 

Extending `DefaultHttpRequestRetryHandler` as a separate class also looks redundant because it is used only as an one instance and only once.

This i decided to extend/implement class inline, but we actually don't need to extend functionality (like add/override methods) thus the body `{}` could be left empty. But this could be confusing from first view. That's why i decided to put one short comment why the body is empty.

Honestly, i don't know how to describe this in a short comment....

service/src/main/java/com/commercetools/util/HttpRequestUtil.java
andrii-kovalenko-ct: I'd also like to know, but i have not found any explicit recommendations for the pool sizes. This value is "inherited" from Unirest, but in Unirest it is also not explained why 200/20 is a good size.
We will be using this numbers as long as we don't face problems with out of memory or out of connections.

service/src/main/java/com/commercetools/util/HttpRequestUtil.java
andrii-kovalenko-ct: I'd also love to do this, but created/returned `CloseableHttpClient` exposes only `execute()` and `close()` methods. The _pool_ and _timeout_ properties are hidden very deep in the implementation.

The only thing i was able to test are short/long timeout responses in the `functionaltests`. 
I'd also like to test pool request size, but for this i have to make a stand alone test, when all funtionaltests are run in parallel doing tens of requests per second

service/src/main/java/com/commercetools/util/HttpRequestUtil.java
heshamMassoud: ok, could you please add this there :)? At least if problems where faced, a future dev would see that there where no specific reason for them in the first place.

service/src/main/java/com/commercetools/util/HttpRequestUtil.java
heshamMassoud: ok, why not make the `buildDefaultConnectionManager` have `package-private` access?

service/src/main/java/com/commercetools/util/HttpRequestUtil.java
heshamMassoud: Thanks for the explanation. As discussed, please add short comment about anonymous extension.

service/src/main/java/com/commercetools/util/HttpRequestUtil.java
andrii-kovalenko-ct: 5e06833e

service/src/main/java/com/commercetools/util/HttpRequestUtil.java
andrii-kovalenko-ct: e76ec38e

service/src/main/java/com/commercetools/util/HttpRequestUtil.java
andrii-kovalenko-ct: does it really have sense to test static stateless method with explicit logic? It is just a dumb   3-lines factory method with straightforward logic and without any branching, looping, extending or inheriting. Actually the only reason this method exists as a separate method is that `HttpClientConnectionManager` doesn't have a builder (like `RequestConfig` which is used in `setDefaultRequestConfig()` above). Hence i carried out `new->set->set` to separate method. Otherwise i would looks same as in `setDefaultRequestConfig()` or `DefaultHttpRequestRetryHandler`.

heshamMassoud: My only suggestion was to have the test as a safeguard against any side effects if someone changes the code in the future. Not to test the actual complexity logic of this method.

andrii-kovalenko-ct: 22fa8000 and be happy :)

Issue #197, opened 2017-12-04
Fix HttpRequestUtil (test) to avoid fail on 408 (timeout) response from httpbin.org


andrii-kovalenko-ct:
http://httpbin.org could sometimes response with _408 Timeout_ response in case of heavy loading. In this case some parallel tests could fail, because _200 OK_ is expected.

There are 2 steps should be done:
  - [x] (fast one for now) - make re-try in `HttpRequestUtilParallelTest` in case of _408_ response
  - [ ] (later) - re-factor `HttpRequestUtil` to make re-try in case of _408_ response.

**Note**: `HttpRequestUtil` already has a retry strategy, but it is applied only in case of socket/connection timeout exception, when the connection is not established or when a response is not received. In case of _408_ this is not handled, because it is a valid in-time response from the server. 
Thus implementation of this case in `HttpRequestUtil` is not simple: `DefaultHttpRequestRetryHandler` should be extended, the response should be parsed and based on the response code - returned or re-tried considering `DefaultHttpRequestRetryHandler#retryCount` value

PR #198, opened 2017-12-04
#197: re-factor HttpRequestUtilParallelTest to handle 408 Timeout res…

andrii-kovalenko-ct:
Addresses #197 issue:
 - implemented re-try strategy for [`HttpRequestUtilParallelTest`](https://github.com/commercetools/commercetools-payone-integration/pull/198/files#diff-68f2f161bb34687ec51123bdbb575269R57) to avoid false failure on _408 Timeout_ response.

Full fix for re-try strategy in `HttpRequestUtil` will come later as a separate solution.

### DIFF
--- a/service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
+++ b/service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
@@ -35,24 +35,28 @@ public class HttpRequestUtilParallelTest {
 
     @Test
     public void executeGetRequest_shouldBeParalleled() throws Exception {
-        makeAndAssertParallelRequests(nThreads, nRequests, createRequestsSupplier(
-                () -> executeGetRequest(HTTP_HTTPBIN_ORG_GET),
-                // in get requests asserted only response status
-                httpResponse -> assertThat(httpResponse.getStatusLine().getStatusCode()).isEqualTo(HttpStatus.SC_OK)));
+        ExceptionalSupplier<HttpResponse> getRequestSupplier = () -> executeGetRequest(HTTP_HTTPBIN_ORG_GET);
+
+        // in get requests asserted only response status
+        ExceptionalConsumer<HttpResponse> getResponseAssertor =
+                httpResponse -> assertThat(httpResponse.getStatusLine().getStatusCode()).isEqualTo(HttpStatus.SC_OK);
+
+        makeAndAssertParallelRequests(nThreads, nRequests, createRequestsValidator(getRequestSupplier, getResponseAssertor));
     }
 
     @Test
     public void executePostRequest_shouldBeParalleled() throws Exception {
-        makeAndAssertParallelRequests(nThreads, nRequests, createRequestsSupplier(
-                () -> executePostRequest(HTTP_HTTPBIN_ORG_POST, asList(
-                        nameValue("xxx", "yyy"),
-                        nameValue("zzz", 11223344))),
-                // in post requests asserted status and body
-                httpResponse -> {
-                    assertThat(httpResponse.getStatusLine().getStatusCode()).isEqualTo(HttpStatus.SC_OK);
-                    assertThat(responseToString(httpResponse)).contains("xxx", "yyy", "zzz", "11223344");
-                }
-        ));
+        ExceptionalSupplier<HttpResponse> postRequestSupplier = () -> executePostRequest(HTTP_HTTPBIN_ORG_POST, asList(
+                nameValue("xxx", "yyy"),
+                nameValue("zzz", 11223344)));
+
+        // in post requests asserted status and body
+        ExceptionalConsumer<HttpResponse> postResponseAssertor = httpResponse -> {
+            assertThat(httpResponse.getStatusLine().getStatusCode()).isEqualTo(HttpStatus.SC_OK);
+            assertThat(responseToString(httpResponse)).contains("xxx", "yyy", "zzz", "11223344");
+        };
+
+        makeAndAssertParallelRequests(nThreads, nRequests, createRequestsValidator(postRequestSupplier, postResponseAssertor));
     }
 
     /**
@@ -68,8 +72,8 @@ public class HttpRequestUtilParallelTest {
      * @param responseAssertConsumer procedure to assert the response obtained from {@code responseSupplier}
      * @return new supplier which when finished successfully - returns <b>1</b>, otherwise - <b>0</b>.
      */
-    private Supplier<Integer> createRequestsSupplier(final ExceptionalSupplier<HttpResponse> responseSupplier,
-                                                     final ExceptionalConsumer<HttpResponse> responseAssertConsumer) {
+    private Supplier<Integer> createRequestsValidator(final ExceptionalSupplier<HttpResponse> responseSupplier,
+                                                      final ExceptionalConsumer<HttpResponse> responseAssertConsumer) {
         return () -> {
             try {
                 int statusCode;
@@ -109,19 +113,19 @@ public class HttpRequestUtilParallelTest {
      * from {@code requestSupplier}. At the end assert that all {@code nRequests} finished successfully, i.e.
      * exception or timeout not happened.
      *
-     * @param nThreads        number of parallel requests/threads
-     * @param nRequests       total number of requests to execute
-     * @param requestSupplier requests generator: generated requests return <b>1</b> if finished successfully or
-     *                        <b>0</b> otherwise
+     * @param nThreads         number of parallel requests/threads
+     * @param nRequests        total number of requests to execute
+     * @param requestValidator requests validator: generates request, validate response  and return <b>1</b> if
+     *                         the request is finished successfully or <b>0</b> otherwise
      * @throws Exception in case of thread pool exceptions
      */
     private void makeAndAssertParallelRequests(final int nThreads, final int nRequests,
-                                               final Supplier<Integer> requestSupplier) throws Exception {
+                                               final Supplier<Integer> requestValidator) throws Exception {
         final ExecutorService newFixedThreadPool = Executors.newFixedThreadPool(nThreads);
 
         final AtomicInteger counter = new AtomicInteger(0);
         for (int i = 0; i < nRequests; i++) {
-            newFixedThreadPool.execute(() -> counter.addAndGet(requestSupplier.get()));
+            newFixedThreadPool.execute(() -> counter.addAndGet(requestValidator.get()));
         }
 
         newFixedThreadPool.shutdown();
--- a/service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
+++ b/service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
@@ -3,8 +3,9 @@ package com.commercetools.util;
 import org.apache.http.HttpResponse;
 import org.apache.http.HttpStatus;
 import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import java.io.IOException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
```diff
@@ -83,7 +141,20 @@ private void makeParallelRequests(final int nThreads, final int requests, final
                 .withFailMessage(format("The execution thread pool was not able to shutdown in time %s msec", maxCriticalPathDurationMsec))
```
heshamMassoud: something maybe not related to this PR, but just a suggestion. Maybe move this threadpool shutdown in a separate method: 

````java
    private void assertThreadPoolIsShutdown(final int nThreads,
                                            final double requestsNumber,
                                            final ExecutorService newFixedThreadPool) throws InterruptedException {
        // length of the longest pipe in the multi thread pipeline,
        // literally it is an maximum integer number of sequential requests in one pipe (thread)
        final int criticalPathLength = (int) Math.ceil(requestsNumber / nThreads);

        // longest expected time of one successful request (even if retried),
        // which may have +RETRY_TIMES attempts additionally to the first (failed) attempt
        final int longestRequestTimeMsec = REQUEST_TIMEOUT * (1 + RETRY_TIMES);

        // await not more than (longestRequestTimeMsec * criticalPathLength) msec with
        // coefficient 1.5 is added to avoid test fails on some lags and threads switching timeouts.
        final long maxCriticalPathDurationMsec = (int) (1.5 * longestRequestTimeMsec * criticalPathLength);
        boolean interrupted = newFixedThreadPool.awaitTermination(maxCriticalPathDurationMsec, TimeUnit.MILLISECONDS);
        assertThat(interrupted)
                .withFailMessage(format("The execution thread pool was not able to shutdown in time %s msec", maxCriticalPathDurationMsec))
                .isTrue();
    }
````

and then in this method only -> 
````java
assertThreadPoolIsShutdown(nThreads, requestsNumber, newFixedThreadPool);
````

File: service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java [hunk 2]
```diff
@@ -24,52 +26,108 @@
...
+                        .withFailMessage(format("Retries over-limit: the post request is executed %d times with timeout response %d",
+                                repeatedTimes, SC_REQUEST_TIMEOUT))
+                        .isLessThanOrEqualTo(RETRY_TIMES);
+
+                responseAssertConsumer.accept(httpResponse);
                 return 1;
-            } catch (IOException e) {
-                throw new RuntimeException("Request exception", e);
+            } catch (Exception e) {
+                LOGGER.error("Request exception: {}", e.toString());
```
heshamMassoud: sometimes you use Strings.format and sometimes you use this string intepolation style. It's better if we are consistent with the style in the same code base. Otherwise, it feels like there is a reason behind it. Unless, there is a reason behind it here which I don't know 😄

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
```diff
@@ -24,52 +26,108 @@
...
+     * from {@code requestSupplier}. At the end assert that all {@code requestsNumber} finished successfully, i.e.
+     * exception or timeout not happened.
+     *
+     * @param nThreads        number of parallel requests/threads
+     * @param requestsNumber  total number of requests to execute
+     * @param requestSupplier requests generator: generated requests return <b>1</b> if finished successfully or
+     *                        <b>0</b> otherwise
+     * @throws Exception in case of thread pool exceptions
+     */
+    private void makeAndAssertParallelRequests(final int nThreads, final int requestsNumber,
```
heshamMassoud: why `nThreads` and `requestsNumber`? 
maybe either
- `nThreads` and `nRequests` 
- `threadsNumber` and `requestsNumber`
?

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
```diff
@@ -83,7 +141,20 @@ private void makeParallelRequests(final int nThreads, final int requests, final
                 .withFailMessage(format("The execution thread pool was not able to shutdown in time %s msec", maxCriticalPathDurationMsec))
                 .isTrue();
 
-        assertThat(counter.get()).isEqualTo(requests);
+        assertThat(counter.get())
+                .withFailMessage(format("Incorrect number of finished requests: expected %d, executed %d. " +
+                        "See the standard error logs (stderr) for exceptions", counter.get(), requestsNumber))
```
heshamMassoud: `Please see`

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
```diff
@@ -24,52 +26,108 @@
...
+                                                     final ExceptionalConsumer<HttpResponse> responseAssertConsumer) {
+        return () -> {
+            try {
+                int statusCode;
+                int repeatedTimes = 0;
+                HttpResponse httpResponse;
+                do {
+                    if (repeatedTimes > 0) {
+                        LOGGER.warn("{} \"Request Timeout\" received from [{}]. Retry {} of {} ...",
+                                SC_REQUEST_TIMEOUT, "http://httpbin.org/", repeatedTimes, RETRY_TIMES);
```
heshamMassoud: I see this string hardcoded 3 times in this file, do you think it makes sense to have a constant for it?

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
heshamMassoud: also here: https://github.com/commercetools/commercetools-payone-integration/pull/198/files#diff-68f2f161bb34687ec51123bdbb575269L28

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
andrii-kovalenko-ct: are you talking about "http://httpbin.org/" ? or some other string?

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
```diff
@@ -24,52 +26,108 @@
...
+                    statusCode = httpResponse.getStatusLine().getStatusCode();
+                    repeatedTimes++;
+                    // sometimes httpbin.org returns 408 (timeout) response. Note, this is not a InterruptedIOException
+                    // or ConnectException, thus DefaultHttpRequestRetryHandler re-try won't handle this case.
+                    // That's why in the tests we use this approach to re-try if the server responded successfully,
+                    // but with "408 timeout" response
+                } while (statusCode == SC_REQUEST_TIMEOUT && repeatedTimes <= RETRY_TIMES);
+
+                assertThat(repeatedTimes)
+                        .withFailMessage(format("Retries over-limit: the post request is executed %d times with timeout response %d",
```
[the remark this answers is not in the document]
andrii-kovalenko-ct: but be careful:
`.withFailMessage()` (or it's origin `.overridingErrorMessage()`) works properly only if you put it **before** assertion condition:
**Correct**:
```java
 assertThat(repeatedTimes)
     .withFailMessage("Retries over-limit")
     .isLessThanOrEqualTo(RETRY_TIMES);
```

**Wrong**:
```java
 assertThat(repeatedTimes)
     .isLessThanOrEqualTo(RETRY_TIMES)
     .withFailMessage("Retries over-limit");
```

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
andrii-kovalenko-ct: yes, there is explicit reason for this:
`slf4j` has it's own string substitution with postponed validation on demand (e.g. if logging actually happens), and this is exactly way recommended by `slf4j` developers:
https://www.slf4j.org/faq.html#logging_performance
https://www.slf4j.org/api/org/slf4j/helpers/MessageFormatter.html

in other cases i use `String.format`, which is actually just a formatting in-place, i.e. executed immediately.

andrii-kovalenko-ct: In the same time i noticed i make redundant `.toString()`, thanks: affc0a07

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
andrii-kovalenko-ct: fc36e546

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
andrii-kovalenko-ct: well, don't think it is necessary right here:
the method is monolithic, not reusable, not public/exposed and fits one screen height. don't think it has much sense to split it .

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
andrii-kovalenko-ct: No, "**See, motherXXX, the logs!**", otherwise you and all your family will die after the ring!

andrii-kovalenko-ct: 455f9258

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
andrii-kovalenko-ct: affc0a07

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
```diff
@@ -24,52 +27,108 @@
...
-                assertThat(httpResponse.getStatusLine().getStatusCode()).isEqualTo(HttpStatus.SC_OK);
-                return 1;
-            } catch (IOException e) {
-                throw new RuntimeException("Request exception", e);
-            }
-        });
+        makeAndAssertParallelRequests(nThreads, nRequests, createRequestsSupplier(
+                () -> executeGetRequest(HTTP_HTTPBIN_ORG_GET),
+                // in get requests asserted only response status
+                httpResponse -> assertThat(httpResponse.getStatusLine().getStatusCode()).isEqualTo(HttpStatus.SC_OK)));
```
LEQADA: It is a bit hard to read. Could we put it in some variable separately?
```
requestSupplier = createRequestsSupplier(
                () -> executeGetRequest(HTTP_HTTPBIN_ORG_GET),
                // in get requests asserted only response status
                httpResponse -> assertThat(httpResponse.getStatusLine().getStatusCode()).isEqualTo(HttpStatus.SC_OK));
makeAndAssertParallelRequests(nThreads, nRequests, requestSupplier);
```

andrii-kovalenko-ct: maybe like this: 32045660

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
heshamMassoud: > are you talking about "http://httpbin.org/" ?

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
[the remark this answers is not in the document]
heshamMassoud: thx

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
heshamMassoud: It was just a recommendation to make it maintainable and easier for future devs.

service/src/test/java/com/commercetools/util/HttpRequestUtilParallelTest.java
andrii-kovalenko-ct: 82455e0e441ff7d447ac13d4d0339766ac5f98fe

PR #201, opened 2018-01-12
Fix integration guide wording

butenkor:


butenkor:
@andrii-kovalenko-ct Thanks. Formatted: https://github.com/commercetools/commercetools-payone-integration/pull/201/commits/7e795de6a6243b1bf22ac16edccaa7673fb7239b

Issue #209, opened 2018-01-31
Fix capture notification processor transaction status


andrii-kovalenko-ct:
- [x] `capture` notification processor must set `Success` status, not `Pending`, otherwise it causes later scheduled messages processors to process new `Charge` transaction, and this could fail for some payments (for example, Klarna doesn't process `Charge` transactions)

 - [x] change transaction status to `Success` if transaction `Charge` already exists

https://github.com/commercetools/commercetools-payone-integration/blob/4c280c3190c6a526f03a777d846f0315a66bee00/service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java#L53-L59

andrii-kovalenko-ct:
closed by #211

PR #211, opened 2018-02-01
209 fix capture transaction status

andrii-kovalenko-ct:
addresses issue #209:

When notification `capture` is received from Payone, we have to set transaction state to `Success` instead of `Pending`, because `capture` means money transferred to the customer.
See [CaptureNotificationProcessor](https://github.com/commercetools/commercetools-payone-integration/pull/211/files#diff-7412916c8bd0808e0e616165c9cc3ed1) class.

Also, if there is no respective `Charge` transaction in the updated payment - [create such a transaction with `Success` state](https://github.com/commercetools/commercetools-payone-integration/pull/211/files#diff-7412916c8bd0808e0e616165c9cc3ed1R74)

[Updated respective tests](https://github.com/commercetools/commercetools-payone-integration/pull/211/files#diff-f0e35d699b64d57c15cf9ef518442b07L88) (only unit, because capture notification never comes on sandbox account).

### DIFF
--- a/service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java
@@ -52,7 +52,7 @@ public class CaptureNotificationProcessor extends NotificationProcessorBase {
 
     /**
      * Add {@link ChangeTransactionState} action to {@code actionsBuilder} if current transaction is still in different
-     * state.
+     * from state in {@code notification}.
      */
     protected ImmutableList.Builder<UpdateAction<Payment>> updateChargeTransactionState(
             @Nonnull final Transaction transaction,
--- a/service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java
@@ -11,8 +11,10 @@ import com.google.common.collect.ImmutableList;
 import io.sphere.sdk.commands.UpdateAction;
 import io.sphere.sdk.payments.*;
 import io.sphere.sdk.payments.commands.updateactions.AddTransaction;
+import io.sphere.sdk.payments.commands.updateactions.ChangeTransactionState;
 import io.sphere.sdk.utils.MoneyImpl;
 
+import javax.annotation.Nonnull;
 import javax.money.MonetaryAmount;
 import java.util.List;
 
@@ -28,7 +30,7 @@ public class CaptureNotificationProcessor extends NotificationProcessorBase {
         super(tenantFactory, tenantConfig, transactionStateResolver);
     }
 
-        @Override
+    @Override
     protected boolean canProcess(final Notification notification) {
         return NotificationAction.CAPTURE.equals(notification.getTxaction());
     }
@@ -36,29 +38,56 @@ public class CaptureNotificationProcessor extends NotificationProcessorBase {
     @Override
     protected ImmutableList<UpdateAction<Payment>> createPaymentUpdates(final Payment payment,
                                                                         final Notification notification) {
-        final ImmutableList.Builder<UpdateAction<Payment>> listBuilder = ImmutableList.builder();
-        listBuilder.addAll(super.createPaymentUpdates(payment, notification));
+        final ImmutableList.Builder<UpdateAction<Payment>> actionsBuilder = ImmutableList.builder();
+        actionsBuilder.addAll(super.createPaymentUpdates(payment, notification));
 
         final List<Transaction> transactions = payment.getTransactions();
         final String sequenceNumber = toSequenceNumber(notification.getSequencenumber());
 
         return findMatchingTransaction(transactions, TransactionType.CHARGE, sequenceNumber)
-                .map(transaction -> listBuilder.build())
-                .orElseGet(() -> {
-                    final MonetaryAmount amount = MoneyImpl.of(notification.getPrice(), notification.getCurrency());
+                .map(transaction -> updateChargeTransactionState(transaction, notification, actionsBuilder))
+                .orElseGet(() -> createChargeTransaction(notification, actionsBuilder, sequenceNumber))
+                .build();
+    }
 
-                    //For all payment methods that use PayOne preauthorization and capture but only one CTP-charge
-                    //we must not create a second charge transaction
-                    //at the moment that is only "BANK_TRANSFER_ADVANCE"
-                    if (ClearingType.PAYONE_VOR != ClearingType.getClearingTypeByCode(notification.getClearingtype())) {
-                    listBuilder.add(AddTransaction.of(TransactionDraftBuilder.of(TransactionType.CHARGE, amount)
-                            .timestamp(toZonedDateTime(notification))
-                            .state(notification.getTransactionStatus().getCtTransactionState())
-                            .interactionId(sequenceNumber)
-                            .build()));
-                    }
+    /**
+     * Add {@link ChangeTransactionState} action to {@code actionsBuilder} if current transaction is still in different
+     * state.
+     */
+    protected ImmutableList.Builder<UpdateAction<Payment>> updateChargeTransactionState(
+            @Nonnull final Transaction transaction,
+            @Nonnull final Notification notification,
+            @Nonnull final ImmutableList.Builder<UpdateAction<Payment>> actionsBuilder) {
+        final TransactionState newTransactionState = notification.getTransactionStatus().getCtTransactionState();
+        if (transaction.getState() != newTransactionState) {
+            actionsBuilder.add(ChangeTransactionState.of(newTransactionState, transaction.getId()));
+        }
 
-                    return listBuilder.build();
-                });
+        return actionsBuilder;
+    }
+
+    /**
+     * For payments except {@code BANK_TRANSFER-ADVANCE} ({@link ClearingType#PAYONE_VOR}) add {@link AddTransaction}
+     * update action with {@link TransactionType#CHARGE} and state equal to
+     * {@link io.sphere.sdk.payments.TransactionState Notification#getTransactionStatus()#getCtTransactionState()}
+     */
+    protected ImmutableList.Builder<UpdateAction<Payment>> createChargeTransaction(
+            @Nonnull final Notification notification,
+            @Nonnull final ImmutableList.Builder<UpdateAction<Payment>> actionsBuilder,
+            @Nonnull final String sequenceNumber) {
+        final MonetaryAmount amount = MoneyImpl.of(notification.getPrice(), notification.getCurrency());
+
+        //For all payment methods that use PayOne preauthorization and capture but only one CTP-charge
+        //we must not create a second charge transaction
+        //at the moment that is only "BANK_TRANSFER_ADVANCE"
+        if (ClearingType.PAYONE_VOR != ClearingType.getClearingTypeByCode(notification.getClearingtype())) {
+            actionsBuilder.add(AddTransaction.of(TransactionDraftBuilder.of(TransactionType.CHARGE, amount)
+                    .timestamp(toZonedDateTime(notification))
+                    .state(notification.getTransactionStatus().getCtTransactionState())
+                    .interactionId(sequenceNumber)
+                    .build()));
+        }
+
+        return actionsBuilder;
     }
 }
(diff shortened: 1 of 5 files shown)


service/src/test/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessorTest.java
```diff
@@ -113,22 +119,26 @@ public void processingPendingNotificationForPendingChargeTransactionDoesNotChang
 
     @Test
     @SuppressWarnings("unchecked")
-    public void processingCompletedNotificationForPendingChargeTransactionDoesNotChangeState() throws Exception {
-        final Payment payment = processingCompletedNotificationForPendingChargeTransactionDoesNotChangeStateWireframe();
+    public void processingCompletedNotificationForPendingChargeTransactionChangesStateToSuccess() throws Exception {
+        final Payment payment = processingCompletedNotificationForPendingChargeTransactionChangesStateWireframe();
         verifyUpdateOrderActions(payment, ORDER_PAYMENT_STATE);
```
ahalberkamp: Maybe you can use this naming convention to make the testname more understandable:
MethodName_StateUnderTest_ExpectedBehavior

andrii-kovalenko-ct: maybe in the future - it requires re-factoring hundreds of functions

service/src/test/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessorTest.java
```diff
@@ -66,12 +63,21 @@ public void setUp() throws Exception {
...
     public void processingPendingNotificationAboutUnknownTransactionAddsChargeTransactionWithStatePending() throws Exception {
+        // note, this case tests un-documented (actually, documented, but not clearly) feature:
+        // it is not clear, could it ever be that txaction=capture has transaction_state=pending.
+        // Doc says [PAYONE Platform Channel Server API, Edition: 2017-08-29, Version: 2.92, Chapter 4.2.2 List of status (transaction_status)]:
+        //   > The parameter “transaction_status” is currently introduced with event-txaction “appointed” only.
+        // but later:
+        //   > Other event-txaction with parameter “transaction_status” may follow (e.g. “paid”, “debit”, ...).
+        //   > The new “txaction” can then be “paid/pending”, “paid/completed”, ... or “failed/completed”.
+        //
+        // So, even if Payone changes API and such case happens - we expect to switch CT transaction state to PENDING, not SUCCESS.
```
heshamMassoud: I am really confused by the comments above.
What does it mean "we expect to switch CT transaction state to PENDING, not SUCCESS."?
Where is it shows SUCCESS in the line of code below?
What does it mean "we expect"? So it's expected to be changed at one point? when? why not changed now?

service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java
```diff
@@ -28,37 +30,64 @@ public CaptureNotificationProcessor(TenantFactory tenantFactory, TenantConfig te
...
-                    if (ClearingType.PAYONE_VOR != ClearingType.getClearingTypeByCode(notification.getClearingtype())) {
-                    listBuilder.add(AddTransaction.of(TransactionDraftBuilder.of(TransactionType.CHARGE, amount)
-                            .timestamp(toZonedDateTime(notification))
-                            .state(TransactionState.PENDING)
-                            .interactionId(sequenceNumber)
-                            .build()));
-                    }
+        //For all payment methods that use PayOne preauthorization and capture but only one CTP-charge
+        //we must not create a second charge transaction
+        //at the moment that is only "BANK_TRANSFER_ADVANCE"
```
heshamMassoud: Is this comment still valid?

heshamMassoud: I can't relate it to the code change in the if condition below..

service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java
```diff
@@ -28,37 +30,64 @@ public CaptureNotificationProcessor(TenantFactory tenantFactory, TenantConfig te
...
-                .orElseGet(() -> {
-                    final MonetaryAmount amount = MoneyImpl.of(notification.getPrice(), notification.getCurrency());
+                .map(transaction -> updateChargeTransactionState(transaction, notification, actionsBuilder))
+                .orElseGet(() -> createChargeTransaction(notification, actionsBuilder, sequenceNumber))
+                .build();
+    }
+
+    /**
+     * Add {@link ChangeTransactionState} action to {@code actionsBuilder} if current transaction is still in different
+     * state.
```
heshamMassoud: what do you mean "still in different state"? : 
1.  different state than what? 
2. "Still"? Does it change at a certain point? what point?

service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java
andrii-kovalenko-ct: the comment is still valid, it is related to this condition:
```java 
ClearingType.PAYONE_VOR != ClearingType.getClearingTypeByCode(notification.getClearingtype())
```

But this might be re-factored in the future cos we also not sure this what we need.

service/src/test/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessorTest.java
andrii-kovalenko-ct: > What does it mean "we expect to switch CT transaction state to PENDING, not SUCCESS."?
> What does it mean "we expect"? So it's expected to be changed at one point? when? why not changed now?

ok, we don't expect, we actually set it to this value. 3de9fe90

> Where is it shows SUCCESS in the line of code below?

the comment says opposite: _switch CT transaction state to **PENDING**_, which is asserted in the function below (_...ChargeTransactionWithState**Pending**_) (i agree the function name is too long,  though it could be re-factored as a separate task/pr)

service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java
andrii-kovalenko-ct: 1. a5310c21

2. The whole class is a _NotificationProcessor_, "still" and "certain point" mean until notification arrived.

service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java
heshamMassoud: @andrii-kovalenko-ct `For all payment methods that use PayOne preauthorization and capture but only one CTP-charge` I still don't see this comment related to the condition..

service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java
```diff
@@ -28,37 +30,64 @@ public CaptureNotificationProcessor(TenantFactory tenantFactory, TenantConfig te
...
-                            .interactionId(sequenceNumber)
-                            .build()));
-                    }
+        //For all payment methods that use PayOne preauthorization and capture but only one CTP-charge
+        //we must not create a second charge transaction
+        //at the moment that is only "BANK_TRANSFER_ADVANCE"
+        if (ClearingType.PAYONE_VOR != ClearingType.getClearingTypeByCode(notification.getClearingtype())) {
+            actionsBuilder.add(AddTransaction.of(TransactionDraftBuilder.of(TransactionType.CHARGE, amount)
+                    .timestamp(toZonedDateTime(notification))
+                    .state(notification.getTransactionStatus().getCtTransactionState())
```
lojzatran: Why don't you set explicitly `TransactionState.SUCCESS`?

service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java
```diff
@@ -28,37 +30,64 @@ public CaptureNotificationProcessor(TenantFactory tenantFactory, TenantConfig te
...
+        actionsBuilder.addAll(super.createPaymentUpdates(payment, notification));
 
         final List<Transaction> transactions = payment.getTransactions();
         final String sequenceNumber = toSequenceNumber(notification.getSequencenumber());
 
         return findMatchingTransaction(transactions, TransactionType.CHARGE, sequenceNumber)
-                .map(transaction -> listBuilder.build())
-                .orElseGet(() -> {
-                    final MonetaryAmount amount = MoneyImpl.of(notification.getPrice(), notification.getCurrency());
+                .map(transaction -> updateChargeTransactionState(transaction, notification, actionsBuilder))
```
lojzatran: Why don't you pass `TransactionState.SUCCESS` instead? It's more explicit what you want in case of `capture`. And here you pass `notification`, but you just use it to get `TransactionState`, so pass `TransactionState` directly instead?

service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java
andrii-kovalenko-ct: i'm trying to write is as a pseudocode:
```java
if (this is a capture notification) { // because this is CaptureNotificationProcessor class
  if (paymentMethod is BANK_TRANSFER_ADVANCE) {
    do nothing (don't add a transaction, because it will be second CHARGE transaction in this payment method)
  } else { // all other payment methods
    add CHARGE transaction
  }
}

Comment (heshamMassoud) [reply service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java]: is `PAYONE_VOR` (in the actual code) the same as `BANK_TRANSFER_ADVANCE` (in the pseudocode)?

Also where is the `if (this is a capture notification) {` in the actual code?

File: service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java (hunk 6 above)
Comment (andrii-kovalenko-ct) [reply service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java]: > Why don't you pass TransactionState.SUCCESS instead? 

Unfortunately, Payone documentation is not clear enough in this point, so it might be that the notification arrives with state `Pending`, but not `Success`. Here even a special test case for this:
https://github.com/commercetools/commercetools-payone-integration/blob/3de9fe90be5fef4bd7feb01b2e2ea78f3518df69/service/src/test/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessorTest.java#L65-L76

Comment (andrii-kovalenko-ct) [reply service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java]: > And here you pass notification, but you just use it to get TransactionState, so pass TransactionState directly instead?

As for passing `notification` or `TransactionState`: i just wanted to simplify the function call, and avoid lambda line look too long like this:
```java
.map(transaction -> updateChargeTransactionState(transaction, notification.getTransactionStatus().getCtTransactionState(), actionsBuilder))
```

Also, this approach is cleaner because we delegate responsibility of transaction update to the function itself: we kind of don't care how `updateChargeTransactionState()` is actually implemented and what properties it requires from the _notification_. Our responsibility is to provide all necessary information. If in the future we have to extend `updateChargeTransactionState()` behavior adding new properties from the notification - we don't have to change the method signature, only the implementation.

service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java
andrii-kovalenko-ct: same as here:
https://github.com/commercetools/commercetools-payone-integration/pull/211#discussion_r165920367

service/src/main/java/com/commercetools/pspadapter/payone/notification/common/CaptureNotificationProcessor.java
andrii-kovalenko-ct: > is PAYONE_VOR (in the actual code) the same as BANK_TRANSFER_ADVANCE (in the pseudocode)?

yes, it's enum value: https://github.com/commercetools/commercetools-payone-integration/blob/d59b48945c6b9777fe3f43ecc9cc491fd17cd52a/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/ClearingType.java#L19

> Also where is the if (this is a capture notification) in the actual code?

like written in the the comment to that line: _because this is **Capture**NotificationProcessor class_, and this means all the logic implemented in this class is executed exclusively in context of `capture` notification processing. I put that `if` into the pseudocode snippet only to explain "_and capture_" words from the comment

heshamMassoud: ok thanks for the explanation.

PR #215, opened 2018-02-28
205 scheduled job test fix

### LINKED ISSUE #205: Re-factor integration tests
- [ ] discuss, estimate and try to migrate from Concordion to normal tests: because of Concordion limitations (match only exact strings, no ranges/lists asserting and so on) some tests are hard to re-factor, support, extend

  - [ ] re-factor order/payment/cart creation like in normal shops workflow, see [Roman's comment](https://github.com/commercetools/commercetools-payone-integration/pull/202#discussion_r164068800) for more details:
    > - generate orderNumber
    > - create payment with given orderNumber
    > - create cart/order with given orderNumber and payment
    > 
    > This would also allow you to skip PaymentUpdateCommand in line 154.

-  make the tests more stable:
  - [x] ~`ScheduledJobShort2` fails very often because the payment is sill not _APPROVED_ after waiting timeout, but later it becomes approved~ (fixed in #215)

andrii-kovalenko-ct:
Make the integration test of scheduled job more stable: [increase waiting timeout](https://github.com/commercetools/commercetools-payone-integration/pull/215/files#diff-7b0194e26c1b0d2b9f1082b0b773db8dR122) because scheduled jobs are sequential and thus may take longer then just 30 seconds to update 2 tenants in a row.

Partially fixes #205 (ScheduledJobShort2 fails very often because the payment is sill not APPROVED after waiting timeout, but later it becomes approved)

### DIFF
--- a/functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
+++ b/functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
@@ -117,10 +117,10 @@ public class ScheduledJobFixtureHelper {
      * scheduled transactions.</li>
      * </ul>
      *
-     * @return 40 (seconds)
+     * @return 59 (seconds)
      */
     public static long waitSecondsTimeout() {
-        return Math.round(30 * 0.95);
+        return Math.round(30 * 1.95);
     }
 
 }
--- a/functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
+++ b/functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
@@ -71,7 +71,7 @@ public class ScheduledJobFixtureHelper {
     public MultiValueResult verifyPaymentIsHandled(String paymentName) throws Exception {
         Payment payment = paymentFixture.fetchPaymentByLegibleName(paymentName);
 
-        logger.info("Current payment [{}] create at {}, now is {}",
+        logger.info("Payment id={} created at {}. The current time now is {}",
                 payment.getId(), payment.getCreatedAt().toInstant(), ZonedDateTime.now().toInstant());
 
         // wait once, if necessary
--- a/functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
+++ b/functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
@@ -9,6 +9,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import specs.response.BasePaymentFixture;
 
+import java.time.Duration;
+import java.time.Instant;
+import java.time.ZonedDateTime;
+
 import static java.util.Optional.ofNullable;
 
 /**
@@ -19,29 +23,30 @@ import static java.util.Optional.ofNullable;
  */
 public class ScheduledJobFixtureHelper {
 
-    private static Logger LOG = LoggerFactory.getLogger(ScheduledJobShortFixture.class);
-
     private final BasePaymentFixture paymentFixture;
+    private final Logger logger;
 
     public ScheduledJobFixtureHelper(BasePaymentFixture paymentFixture) {
         this.paymentFixture = paymentFixture;
+        this.logger = LoggerFactory.getLogger(paymentFixture.getClass());
     }
 
     /**
      * Create a payment, but don't handle it
-     * @param paymentName      test payment alias
-     * @param paymentMethod    payone payment method name
-     * @param transactionType  payone payment transaction type
-     * @param centAmount       payment amount
-     * @param currencyCode     payment currency
+     *
+     * @param paymentName     test payment alias
+     * @param paymentMethod   payone payment method name
+     * @param transactionType payone payment transaction type
+     * @param centAmount      payment amount
+     * @param currencyCode    payment currency
      * @return map of created payment properties
      * @throws Exception any exceptions in the tests
      */
     public MultiValueResult createPayment(String paymentName,
-                                                 String paymentMethod,
-                                                 String transactionType,
-                                                 String centAmount,
-                                                 String currencyCode) throws Exception {
+                                          String paymentMethod,
+                                          String transactionType,
+                                          String centAmount,
+                                          String currencyCode) throws Exception {
 
         Payment payment = paymentFixture.createAndSavePayment(paymentName, paymentMethod, transactionType, centAmount, currencyCode);
 
@@ -58,6 +63,7 @@ public class ScheduledJobFixtureHelper {
 
     /**
      * Verify the {@code paymentName} is automatically handled by (short) scheduled task.
+     *
      * @param paymentName payment alias
      * @return map of the payment properties after waiting scheduled task.
      * @throws Exception any exceptions in the tests
@@ -65,18 +71,23 @@ public class ScheduledJobFixtureHelper {
     public MultiValueResult verifyPaymentIsHandled(String paymentName) throws Exception {
         Payment payment = paymentFixture.fetchPaymentByLegibleName(paymentName);
 
+        logger.info("Current payment [{}] create at {}, now is {}",
+                payment.getId(), payment.getCreatedAt().toInstant(), ZonedDateTime.now().toInstant());
+
         // wait once, if necessary
         if (payment.getPaymentStatus() == null || StringUtils.isBlank(payment.getPaymentStatus().getInterfaceCode())) {
 
             // how much time to wait till the next scheduled job is definitely started
-            long secondsToWait = payment.getCreatedAt()
-                    .plusSeconds(waitSecondsTimeout())
-                    .minusSeconds(System.currentTimeMillis() / 1000).toEpochSecond();
+            long secondsToWait = waitSecondsTimeout() -
+                    Duration.between(payment.getCreatedAt(), ZonedDateTime.now()).getSeconds();
 
             if (secondsToWait > 0) {
-                LOG.info("Wait up to {} seconds for scheduled job to run", secondsToWait);
+                logger.info("Wait up to {} seconds for scheduled job to run", secondsToWait);
                 Thread.sleep(secondsToWait * 1000);
             }
+
+            logger.info("{} seconds passed after the payment creation - verify the payment now",
+                    Duration.between(payment.getCreatedAt(), ZonedDateTime.now()).getSeconds());
 
             payment = paymentFixture.fetchPaymentByLegibleName(paymentName);
         }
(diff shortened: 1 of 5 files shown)


functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
```diff
@@ -19,29 +23,30 @@
...
 public class ScheduledJobFixtureHelper {
 
-    private static Logger LOG = LoggerFactory.getLogger(ScheduledJobShortFixture.class);
-
     private final BasePaymentFixture paymentFixture;
+    private final Logger logger;
 
     public ScheduledJobFixtureHelper(BasePaymentFixture paymentFixture) {
         this.paymentFixture = paymentFixture;
+        this.logger = LoggerFactory.getLogger(paymentFixture.getClass());
```
heshamMassoud: why change the logger's name to the class of the `paymentFixture` instance class instead of this class' (`ScheduledJobFixtureHelper`) name? wouldn't it be confusing while debugging/checking the logs?

functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
```diff
@@ -58,26 +63,32 @@ public MultiValueResult createPayment(String paymentName,
...
      * Verify the {@code paymentName} is automatically handled by (short) scheduled task.
+     *
      * @param paymentName payment alias
      * @return map of the payment properties after waiting scheduled task.
      * @throws Exception any exceptions in the tests
      */
     public MultiValueResult verifyPaymentIsHandled(String paymentName) throws Exception {
         Payment payment = paymentFixture.fetchPaymentByLegibleName(paymentName);
 
+        logger.info("Current payment [{}] create at {}, now is {}",
```
heshamMassoud: So a log message could be as folllows: `"Current payment [123] create at 12.01.2018 10:10:10, now is 01.03.2018 08:00:00" `

I see several problems with such message:
1. Why "Current"? doesn't make sense when seen in the logs, I guess.
2. `"Current payment [123]"`, when I see that, I don't know what is "123", is it key? is it id? is it name?.
3. The sentence has wrong grammar, which could lead to a different meaning: payment created at x, is now created at y. (even if it doesn't make sense, but still having ", now is x" conveys that something changed from x -> y.

So I would recommend reformatting it to:
`"Payment with id:[123] is created at 12.01.2018 10:10:10. The current time now is 01.03.2018 08:00:00" `

functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
```diff
@@ -86,25 +97,30 @@ public MultiValueResult verifyPaymentIsHandled(String paymentName) throws Except
...
+     * scheduled transactions.</li>
      * </ul>
      *
-     * @return 40 (seconds)
+     * @return 59 (seconds)
      */
-    public static int waitSecondsTimeout() {
-        return 40;
+    public static long waitSecondsTimeout() {
+        return Math.round(30 * 1.95);
```
heshamMassoud: any explanation for the `"almost double"` mentioned in the Javadoc/`1.95`? and why not just double/`2`?

functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
```diff
@@ -58,26 +63,32 @@ public MultiValueResult createPayment(String paymentName,
...
+            long secondsToWait = waitSecondsTimeout() -
+                    Duration.between(payment.getCreatedAt(), ZonedDateTime.now()).getSeconds();
 
             if (secondsToWait > 0) {
-                LOG.info("Wait up to {} seconds for scheduled job to run", secondsToWait);
+                logger.info("Wait up to {} seconds for scheduled job to run", secondsToWait);
                 Thread.sleep(secondsToWait * 1000);
             }
 
+            logger.info("{} seconds passed after the payment creation - verify the payment now",
```
heshamMassoud: what does it mean `" - verify the payment now"`? just check if it exists? or is there something else to do to verify it?

functionaltests/src/test/resources/specs/scheduler/ScheduledJobShort.html
```diff
@@ -76,7 +76,9 @@ <h4>Wait <span c:echo="waitSecondsTimeout()"></span> seconds and verify the paym
             <th c:assertEquals="#verifyResult.paymentStatus">Payment Status</th>
             <th c:assertEquals="#verifyResult.interfaceInteractionsSize">Interface Interactions</th>
             <th c:assertEquals="#verifyResult.transactionsSize">Transactions</th>
-            <th c:echo="#verifyResult.lastModifiedAt">Interface Interactions</th>
+            <th c:echo="#verifyResult.createdAt">Created At (FIO)</th>
```
heshamMassoud: nitpick: Usually (words with less than 4 letters e.g. most prepositions) should not be capitalised in titles ;) https://english.stackexchange.com/questions/14/which-words-in-a-title-should-be-capitalized

this is not imp. tho.

functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
andrii-kovalenko-ct: exactly to avoid confusion: `ScheduledJobFixtureHelper` is re-used in several classes, but i want to know which actual class (`ScheduledJobShortFixture` or `ScheduledJobShortFixture2`) is logging right now.

functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
andrii-kovalenko-ct: > any explanation for the "almost double"

from the javadoc:

> to let the job complete if any lags occurred. Also, this is important since the scheduled jobs are sequential (e.g. process every tenant queues one-by-one), hence second tenant transactions processing is blocked in the queue until the first tenant processes all transactions.

> why not just double

To avoid the case when the job is executed twice. We kind of want to verify the scheduled job is executed in time, even with lags.

In any case - there are no strict requirements for this test.

functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
andrii-kovalenko-ct: "verify" means assert all the values in this fixture:
https://github.com/commercetools/commercetools-payone-integration/tree/be93c1add566032c3906d86879fc8238f9993ab8/functionaltests/src/test/resources/specs/scheduler

functionaltests/src/test/resources/specs/scheduler/ScheduledJobShort.html
andrii-kovalenko-ct: a284d4d4

functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
andrii-kovalenko-ct: 5da23f16

functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
heshamMassoud: ok

functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
heshamMassoud: Maybe add this info in the log message? because first thing that came to my mind was to check the payment in impex playground

functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
heshamMassoud: then please at least add this comment "To avoid the case when the job is executed twice. We kind of want to verify the scheduled job is executed in time, even with lags." in the javadoc.

andrii-kovalenko-ct: actually, this won't avoid hundred percent. It's just a wish. This test doesn't have such strict requirements, thus i don't want to specify so strict which time frames should be used: i want to leave it flexible for future if we have to adjust the tests again. If i put such comment - in the future it will be a new discussion: is it allowed to break this "avoid executed twice" or not. 
The test is still in "fuzzy" mode and could be changed.

Also, the feature will be removed at all soon - doesn't have sense to spend a lot of time clarifying the tests requirements.

So, i'd better leave it like this.

functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
andrii-kovalenko-ct: no, log messages are only to see them on Travis build and have some "checkpoints" to understand when and why the test is failing.
Actually, all the (automation) tests about verifying/asserting objects states after some actions, doesn't make sense to verify on impex. If in some rare cases this is required - then yes, i'd put a special comment "verify manually on impex playground"

functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
heshamMassoud: ok, do you think it makes sense to add this information to the Javadoc? that this "won't avoid the problem 100%", and that the values might be needed to adjust again" ?

functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
heshamMassoud: ok, maybe I am not familiar with the testing setup of this concordion. But for me, "verify the payment now" doesn't help me where/how to verify, to be honest.

andrii-kovalenko-ct: it will be done in [`ScheduledJobShort`](https://github.com/commercetools/commercetools-payone-integration/blob/476b3a23d02c64e384dbb6c1f2df3720f6d2aed3/functionaltests/src/test/resources/specs/scheduler/ScheduledJobShort.html)

functionaltests/src/test/java/specs/scheduler/ScheduledJobFixtureHelper.java
andrii-kovalenko-ct: no, i don't think it is necessary in this certain case, especially in view to the fact _the feature will be removed at all soon_

Issue #213, opened 2018-02-21
Re-factor HttpRequestUtilTimeoutTest to avoid dependency on slowwly resource


andrii-kovalenko-ct:
Looks like http://slowwly.robertomurray.co.uk/ is permanently down permanently, so we have to either:
  - [ ] find another similar resource which allows delayed requests
  - [ ] (better) implement some internal way to perform such tests (maybe raise local web service for the test)

PR #216, opened 2018-03-01
#213: fix HttpRequestUtilTimeoutTest to use own mock HTTP service

andrii-kovalenko-ct:
Addressing #213:

Before the test relied on remote http://slowwly.robertomurray.co.uk/ resource, but now the service is down, so own dedicated HTTP service is created for the tests:

  - [Sun's `HttpServer`](https://github.com/commercetools/commercetools-payone-integration/pull/216/files#diff-554579bde49a7c01276233a633d317b1R62) instance is created on random port
  - it [just _sleeps_ the timeout](https://github.com/commercetools/commercetools-payone-integration/pull/216/files#diff-554579bde49a7c01276233a633d317b1R87) specified in the URL request before answering _200 OK_


service/src/test/java/com/commercetools/util/HttpRequestUtilTimeoutTest.java
```diff
@@ -1,24 +1,112 @@
...
+     * HttpServer</a>
+     */
+    private HttpServer server;
+
+
+    /**
+     * Create mock http service which responses after some <i>delay</i>, specified in the the GET query.
      */
-    private static final String SLOWWLY_URL_PATTERN = "http://slowwly.robertomurray.co.uk/delay/%d/url/http://google.com";
+    @Before
```
lojzatran: Why not `@BeforeClass`?

andrii-kovalenko-ct: Because looks like the service is not completely idempotent and stateless, thus i have to make it clean and fresh for every test case.

service/src/test/java/com/commercetools/util/HttpRequestUtilTimeoutTest.java
```diff
@@ -1,24 +1,112 @@
...
+            httpExchange.getResponseBody().close();
+        });
+
+        server.start();
+    }
+
+    @After
+    public void tearDown() {
+        if (server != null) {
+            server.removeContext("/");
```
heshamMassoud: why is it needed? If you gonna stop it anyways after?

service/src/test/java/com/commercetools/util/HttpRequestUtilTimeoutTest.java
```diff
@@ -1,24 +1,112 @@
...
+            server.stop(1);
+            server = null;
```
heshamMassoud: why do you nullify? when you nullify on setup?

service/src/test/java/com/commercetools/util/HttpRequestUtilTimeoutTest.java
```diff
@@ -1,24 +1,112 @@
...
+        });
+
+        server.start();
+    }
+
+    @After
+    public void tearDown() {
+        if (server != null) {
+            server.removeContext("/");
+            server.stop(1);
```
heshamMassoud: why 1 seconds as delay? and not 0? or more than 1?

service/src/test/java/com/commercetools/util/HttpRequestUtilTimeoutTest.java
```diff
@@ -1,24 +1,112 @@
...
+     * Create mock http service which responses after some <i>delay</i>, specified in the the GET query.
      */
-    private static final String SLOWWLY_URL_PATTERN = "http://slowwly.robertomurray.co.uk/delay/%d/url/http://google.com";
+    @Before
+    public void setUp() {
+        int retries = 3;
+        server = null;
+
+        // run the server on random port
+        do {
```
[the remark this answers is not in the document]
andrii-kovalenko-ct: this one is exactly a canonical example: if smth must be executed at least once - `do-while` is a good approach.
In this case we try to start service and definitely we want to try to start it at least once:

https://stackoverflow.com/a/3347010/907576
https://stackoverflow.com/a/224075/907576

(also, why do you think `do-while` is not a _normal_ statement? Just open your mind :smirk:)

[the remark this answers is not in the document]
heshamMassoud: >  if smth must be executed at least once - do-while is a good approach.

@andrii-kovalenko-ct 
But it will be executed at least once also, with a normal while :) 

I just see that the condition will always be true for the first time from your code:
````java
int retries = 3;
while(retries > 0)
````
 that's why I see no difference :) That's why I ask, in case there is something I didn't notice ;)

I definitely think a `do-while` is normal, but I used "normal" to refer to the standard while 😉

service/src/test/java/com/commercetools/util/HttpRequestUtilTimeoutTest.java
andrii-kovalenko-ct: i'm still not 100% sure how it works under the hood, but looks like the server instance is a global, and will be re-used on next instantiation. So, to be 100% sure we cleaned everything.

service/src/test/java/com/commercetools/util/HttpRequestUtilTimeoutTest.java
andrii-kovalenko-ct: just got is from some example, maybe 1 is redundant:
baede087

service/src/test/java/com/commercetools/util/HttpRequestUtilTimeoutTest.java
heshamMassoud: for me, It feels like there is a redundancy here. If anyways you gonna stop and also nullify. Then this doesn't really make sense to be there at all. Actually, it's confusing.

service/src/test/java/com/commercetools/util/HttpRequestUtilTimeoutTest.java
heshamMassoud: I would even remove it and keep it at one place. Either at start up or teardown. Since there is "no reason".

service/src/test/java/com/commercetools/util/HttpRequestUtilTimeoutTest.java
andrii-kovalenko-ct: c3c255ea

service/src/test/java/com/commercetools/util/HttpRequestUtilTimeoutTest.java
andrii-kovalenko-ct: c3c255ea

PR #224, opened 2018-08-20
Add ClearSecuredValuesSerializer annotation to all sensitive bean fields

LEQADA:
## Description
Customer related data shouldn't be logged. Therefore we need to provide a switch in the configuration which allows to anonymise/de-anonymise those data.
This switch implemented as an environment variable `HIDE_CUSTOMER_PERSONAL_DATA` which can be in 2 states: `"true"` OR `"false"`.
Default state is `"true"`. So if the environment variable is not provided then customer related data will be replaced by `<HIDDEN>` placeholder.

### DIFF
--- a/service/src/main/java/com/commercetools/pspadapter/payone/util/ClearSecuredValuesSerializer.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/util/ClearSecuredValuesSerializer.java
@@ -20,7 +20,7 @@ import static com.commercetools.pspadapter.payone.config.PropertyProvider.HIDE_C
  * {@code Apply} accepts 1 boolean parameter. If the parameter is {@code true} then the value will be hidden.
  * If the parameter is {@code false} or was not provided then the code will check environment variable
  * {@code HIDE_CUSTOMER_PERSONAL_DATA}. If the value of the environment variable is {@code true} then the
- * field will be hidden
+ * field will be hidden. If the value of environment variable is not provided then it will be counted as {@code true}
  *
  * @author fhaertig
  * @since 19.04.16
--- a/service/src/main/java/com/commercetools/pspadapter/payone/util/ClearSecuredValuesSerializer.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/util/ClearSecuredValuesSerializer.java
@@ -1,5 +1,6 @@
 package com.commercetools.pspadapter.payone.util;
 
+import com.commercetools.pspadapter.payone.config.PropertyProvider;
 import com.fasterxml.jackson.core.JsonGenerator;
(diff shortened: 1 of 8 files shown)


LEQADA:
@butenkor you mean this one?
https://github.com/commercetools/commercetools-payone-integration/pull/224/files#diff-890fee622c7936209141adf0aa0b2988

service/src/test/java/com/commercetools/pspadapter/payone/domain/payone/model/common/AuthorizationRequestTest.java
```diff
@@ -0,0 +1,270 @@
...
+
+    @Test
+    public void createsMapWithHiddenSecrets() {
+        assertThat(request.toStringMap(true)).containsOnly(
+                entry("narrative_text", ClearSecuredValuesSerializer.PLACEHOLDER),
+                entry("param", ClearSecuredValuesSerializer.PLACEHOLDER),
+                entry("key", ClearSecuredValuesSerializer.PLACEHOLDER),
+                entry("clearingtype", clearingType),
+                entry("errorurl", errorUrl),
+                entry("vatid", vatId),
```
lojzatran: What about this VAT ID? Isn't it also personal info?

LEQADA: @floriansattler ?

LEQADA:
@lojzatran there is https://github.com/commercetools/commercetools-payone-integration/blob/0997157406be6bf16bc77eeac79256c26fa17367/service/src/test/java/com/commercetools/pspadapter/payone/domain/payone/model/banktransfer/BankTransferRequestTest.java

service/src/main/java/com/commercetools/pspadapter/payone/util/ClearSecuredValuesSerializer.java
```diff
@@ -31,7 +40,11 @@ private ClearSecuredValuesSerializer(final BeanProperty property) {
 
     @Override
     public void serialize(final String value, final JsonGenerator gen, final SerializerProvider serializers) throws IOException, JsonProcessingException {
```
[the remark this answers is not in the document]
LEQADA: https://github.com/commercetools/commercetools-payone-integration/pull/224/commits/0997157406be6bf16bc77eeac79256c26fa17367

service/src/test/java/com/commercetools/pspadapter/payone/domain/payone/model/common/AuthorizationRequestTest.java
floriansattler: It`s a public info, as it is from companies.

PR #225, opened 2018-12-31
CTPI-197 - update dependencies

LEQADA:


butenkor:
@lojzatran Did you consider also to update other dependencies? If not major then at least bug fix versions.

lojzatran:
So unfortunately I couldn't update guava. In the newer versions, the test `CashinadvanceFixture` couldn't finish and it got stuck as in this build: https://travis-ci.org/commercetools/commercetools-payone-integration/builds/560887012

PR #230, opened 2019-03-26
Update chrome driver

ahalberkamp:
In this pr, it was tried to "stabilize" the build process. For doing this the following was done:

1) Updating the webdriver to version
2) Executing the fixtures for sofortüberweisung sequentiell (instead of parellel).

### DIFF
--- a/functionaltests/src/test/java/specs/paymentmethods/sofortueberweisung/ChargeImmediatelyFixture.java
+++ b/functionaltests/src/test/java/specs/paymentmethods/sofortueberweisung/ChargeImmediatelyFixture.java
@@ -152,7 +152,7 @@ public class ChargeImmediatelyFixture extends BaseNotifiablePaymentFixture {
     public boolean executeRedirectForPayments(final String paymentNames) throws ExecutionException {
         final Collection<String> paymentNamesList = ImmutableList.copyOf(thePaymentNamesSplitter.split(paymentNames));
 
-        // run all 3 payments approval in parallel, aka 3 different sessions
+        // run all 3 payments approval, aka 3 different sessions
         // and collect successfully approved redirect URLs
         successUrlForPayment = paymentNamesList.stream()
                 .map(this::approvePaymentAsCustomer)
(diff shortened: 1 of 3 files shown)


functionaltests/src/test/java/specs/paymentmethods/sofortueberweisung/ChargeImmediatelyFixture.java
```diff
@@ -154,7 +154,7 @@ public boolean executeRedirectForPayments(final String paymentNames) throws Exec
 
         // run all 3 payments approval in parallel, aka 3 different sessions
         // and collect successfully approved redirect URLs
```
butenkor: You have to change comment accordingly and please also explain in the comment why it is not in parallel anymore.

lojzatran:
The tests are failing now.

functionaltests/src/test/java/specs/paymentmethods/sofortueberweisung/ChargeImmediatelyFixture.java
lojzatran: I tried to put back `parallel`, but the tests fails, so this fix is correct. @ahalberkamp pls have a look and answer Roman's comment.

butenkor: @lojzatran yes fix is as described in PR description to reduce parallel execution but my request was to just align the java code comment to the change.

functionaltests/src/test/java/util/CustomWebDriver.java
```diff
@@ -14,7 +14,7 @@
     private ChromeDriver driver;
 
     CustomWebDriver() {
-        String currentDriver = SystemUtils.IS_OS_MAC ? "chromedriver_ios_2.40" : "chromedriver_unix_2.40";
+        String currentDriver = SystemUtils.IS_OS_MAC ? "chromedriver_mac_75" : "chromedriver_linux_75";
```
butenkor: Why not latest version which is 76.0.3809.68 right now? See: http://chromedriver.chromium.org/downloads

lojzatran: Because it's not stable version, but only beta. And we're testing with stable version. https://chromereleases.googleblog.com/

functionaltests/src/test/java/specs/paymentmethods/sofortueberweisung/ChargeImmediatelyFixture.java
lojzatran: https://github.com/commercetools/commercetools-payone-integration/pull/230/commits/f83ef63d7885699b6a7555d3767470734977171e

PR #234, opened 2019-09-04
Log exception, when payment cannot be found.

ahalberkamp:
In this pr an exception was added to the error-log to simplify debugging


service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
```diff
@@ -83,8 +83,10 @@ public PaymentHandleResult handlePayment(@Nonnull final String paymentId) {
         }
     }
 
-    private PaymentHandleResult handleNotFoundException(@Nonnull String paymentId) {
-        final String body = format("Could not process payment with ID [%s]: order or cart not found", paymentId);
+     private PaymentHandleResult handleNotFoundException(@Nonnull String paymentId, @Nonnull Exception exception ) {
+        final String body = format("Could not process payment with ID [%s], paymentId");
```
butenkor:
```suggestion
        final String body = format("Error on processing payment with ID [%s] as payment or cart could not be found", paymentId);
```

PR #236, opened 2019-09-19
CAR-2726 remove ctp payment checks from health endpoint

ahmetoz:
The current implementation is causing some problems if ctp is failing with `504` or `500` status. So k8s restarting pod because of the platform issue, this leads some unexpected restarts.

with this PR, if the health endpoint is working, it will return HTTP ok (200) when it called by k8s  liveness/readiness probes. 

JIRA ticket: https://jira.commercetools.com/browse/CAR-2726

### DIFF
--- a/service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java
@@ -22,9 +22,12 @@ import spark.utils.CollectionUtils;
 import javax.annotation.Nonnull;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import static com.commercetools.pspadapter.payone.util.PayoneConstants.PAYONE;
 import static io.sphere.sdk.json.SphereJsonUtils.toJsonString;
@@ -40,9 +43,8 @@ import static java.util.stream.Collectors.toMap;
 public class IntegrationService {
 
     public static final Logger LOG = LoggerFactory.getLogger(IntegrationService.class);
-    static final int ERROR_STATUS = HttpStatus.SERVICE_UNAVAILABLE_503;
     static final int SUCCESS_STATUS = HttpStatus.OK_200;
-    static final String STATUS_KEY = "status";
+    private static final String STATUS_KEY = "status";
     static final String TENANTS_KEY = "tenants";
     static final String APPLICATION_INFO_KEY = "applicationInfo";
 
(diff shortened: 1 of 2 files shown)


service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java
```diff
@@ -182,29 +178,15 @@ public int port() {
...
                 "version", serviceConfig.getApplicationVersion(),
                 "title", serviceConfig.getApplicationName());
-        Map<String, CompletionStage<Integer>> tenantMap = checkTenantStatuses(tenants);
-        //resolve all completable stages
-        listOfFuturesToFutureOfList(new ArrayList<>(tenantMap.values())).join();
-        //unpack completable features
-        Map<String, Integer> statusMap = tenantMap.keySet().stream()
-                .collect(toMap(tenant -> tenant, tenant -> tenantMap.get(tenant).toCompletableFuture().join()));
+
+        final Map<String, Integer> tenantStatutes = new HashMap<>();
```
LEQADA: Why do we need these statuses? We are not checking anything tenant related, right?

service/src/test/java/com/commercetools/pspadapter/payone/IntegrationServiceTest.java
```diff
@@ -72,52 +63,20 @@ public void tearDown() throws Exception {
 
     @Test
     public void createHealthResponse_tenantConfigIsWorking_shouldReturnOkStatus() {
```
LEQADA: We are actually not checking tenant configs anymore. So I'd remove it from the test name.

service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java
ahmetoz: > We are not checking anything tenant related, right?
Yes.

> Why do we need these statuses? 
I also don't know why we needed those checks, I guess it's a multitenant based implementation, so I did not want to change the current response logic of health check. 
But we could, wdyt ? should we change it?

LEQADA: We needed it because we were checking the CTP connection for each tenant's project. Since we're not checking CTP connection anymore, I'd say we don't need these statuses.

ahmetoz: Makes sense 👍

ahmetoz: f558a77

service/src/test/java/com/commercetools/pspadapter/payone/IntegrationServiceTest.java
ahmetoz: https://github.com/commercetools/commercetools-payone-integration/pull/236/commits/f558a77fbeb34001167daddefd716ccc8224e0eb#diff-93bd210e428df4a1fb1d0a0c19798021

PR #238, opened 2019-10-17
Fix paydirekt test

ahalberkamp:
In this pr, the test for paydirekt was fixed. The fix was required for the selenium test, because the UI of paydirekt was changed.

### DIFF
--- a/functionaltests/src/test/resources/specs/response/HealthResponse.html
+++ b/functionaltests/src/test/resources/specs/response/HealthResponse.html
@@ -15,8 +15,6 @@ Test we get correct response status, mime-type and message from
             <th c:assertEquals="#result.statusCode">Status Code</th>
             <th c:assertEquals="#result.mimeType">Mime Type</th>
             <th c:assertEquals="#result.bodyStatus">Body status({&quot;status&quot;:200})</th>
-            <th c:echo="#result.bodyTenants">Tenants (info)</th>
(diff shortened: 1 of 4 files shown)


ahalberkamp:
will do

functionaltests/src/test/resources/specs/response/HealthResponse.html
```diff
@@ -15,8 +15,6 @@ <h1>Response from <code>/health</code> URL</h1>
             <th c:assertEquals="#result.statusCode">Status Code</th>
             <th c:assertEquals="#result.mimeType">Mime Type</th>
             <th c:assertEquals="#result.bodyStatus">Body status({&quot;status&quot;:200})</th>
-            <th c:echo="#result.bodyTenants">Tenants (info)</th>
-            <th c:assertEquals="#result.bodyTenantsSize">Tenants list size</th>
```
butenkor: @ahalberkamp Why it has been removed? How is this related to the web test fix?

ahalberkamp: Because the new health-check response doesn`t contain any information about tenants anymore... see:https://github.com/commercetools/commercetools-payone-integration/commit/4c74c0f5924920385bbb798243640f8dcb330980#diff-a684bbe1ce55a8f4dea792d69dd8397aL185

And because of the missing tenant-information the functionaltest from above failed.

PR #241, opened 2019-12-17
Stackdriver logging

heshamMassoud:
To be pair reviewed with @ahalberkamp


service/src/main/resources/logback.xml
```diff
@@ -1,14 +1,30 @@
...
-    <conversionRule conversionWord="coloredLevel" converterClass="play.api.libs.logback.ColoredLevel" />
-
-    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
-        <encoder>
-            <pattern>%date{HH:mm:ss.SSS} %highlight([%-5level]) %cyan(%logger{16}) [%thread] - %message%n%xException{10}</pattern>
+    <appender name="JSON-STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder class="net.logstash.logback.encoder.LoggingEventCompositeJsonEncoder">
+            <providers>
+                <timestamp>
+                    <fieldName>timestamp</fieldName>
```
heshamMassoud: not needed.

service/src/main/resources/logback.xml
```diff
@@ -1,14 +1,30 @@
 <configuration>
-    <conversionRule conversionWord="coloredLevel" converterClass="play.api.libs.logback.ColoredLevel" />
-
-    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
-        <encoder>
-            <pattern>%date{HH:mm:ss.SSS} %highlight([%-5level]) %cyan(%logger{16}) [%thread] - %message%n%xException{10}</pattern>
+    <appender name="JSON-STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder class="net.logstash.logback.encoder.LoggingEventCompositeJsonEncoder">
```
heshamMassoud: Maybe `LogstashEncoder` is enough for our needs now.

service/src/main/resources/logback.xml
heshamMassoud: actually needed to remove the output name with `@timestamp` -> `timestamp`
https://github.com/logstash/logstash-logback-encoder#standard-fields

service/src/main/resources/logback.xml
heshamMassoud: Donesn't hurt now + more flexibility later on when we use MDC for correlation id in the future.

PR #257, opened 2019-12-20
Remove redundant logs

heshamMassoud:
- Remove redundant logs
- Use tenant name as marker instead of logger name.
- Always add the exception, if there is, in the log stack trace.

### DIFF
--- a/service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
@@ -80,7 +80,7 @@ public class PaymentHandler {
         }
     }
 
-    private PaymentHandleResult processPayment(@Nonnull final String paymentId) {
+    private PaymentHandleResult processPayment(@Nonnull final String paymentId) throws ConcurrentModificationException {
         final PaymentWithCartLike paymentWithCartLike =
             commercetoolsQueryExecutor
                 .getPaymentWithCartLike(paymentId);
--- a/service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
@@ -52,31 +52,30 @@ public class PaymentHandler {
      */
     public PaymentHandleResult handlePayment(@Nonnull final String paymentId) {
         try {
+            ConcurrentModificationException lastConcurrentModificationException = null;
             for (int i = 0; i < RETRIES_LIMIT; i++) {
                 try {
-                    final PaymentWithCartLike paymentWithCartLike = commercetoolsQueryExecutor.getPaymentWithCartLike(paymentId);
-                    String paymentInterface = paymentWithCartLike.getPayment().getPaymentMethodInfo().getPaymentInterface();
+                    final PaymentWithCartLike paymentWithCartLike = commercetoolsQueryExecutor
+                        .getPaymentWithCartLike(paymentId);
+                    String paymentInterface = paymentWithCartLike.getPayment().getPaymentMethodInfo()
+                                                                 .getPaymentInterface();
                     if (!payoneInterfaceName.equals(paymentInterface)) {
-                        logger.warn(tenantNameKeyValue, "Wrong payment interface name: expected [{}], found [{}]", payoneInterfaceName, paymentInterface);
+                        logger.warn(tenantNameKeyValue, "Wrong payment interface name: expected [{}], found [{}]",
+                            payoneInterfaceName, paymentInterface);
                         return new PaymentHandleResult(HttpStatusCode.BAD_REQUEST_400);
                     }
 
                     paymentDispatcher.dispatchPayment(paymentWithCartLike);
                     return new PaymentHandleResult(HttpStatusCode.OK_200);
-                } catch (final ConcurrentModificationException cme) {
-                    logger.warn(tenantNameKeyValue,
-                        "Exception on dispatchPayment with id [{}]. Retry {} of {}", paymentId, i + 1, RETRIES_LIMIT,
-                        cme);
+                } catch (final ConcurrentModificationException concurrentModificationException) {
+                    lastConcurrentModificationException = concurrentModificationException;
                     Thread.sleep(RETRY_DELAY);
                 }
             }
-
-            logger.error(
-                tenantNameKeyValue,
-                "The payment [{}] couldn't be processed after {} retries", paymentId, RETRIES_LIMIT);
-            return new PaymentHandleResult(HttpStatusCode.ACCEPTED_202,
-                    format("The payment couldn't be processed after %s retries", RETRIES_LIMIT));
-
+            throw lastConcurrentModificationException;
+        }
+        catch (final ConcurrentModificationException concurrentModificationException) {
+            return handleConcurrentModificationException(paymentId, concurrentModificationException);
         } catch (final NotFoundException | NoCartLikeFoundException e) {
             return handleNotFoundException(paymentId, e);
         } catch (final ErrorResponseException e) {
@@ -86,7 +85,17 @@ public class PaymentHandler {
         }
     }
 
-     private PaymentHandleResult handleNotFoundException(@Nonnull String paymentId, @Nonnull Exception exception ) {
+    private PaymentHandleResult handleConcurrentModificationException(
+        @Nonnull final String paymentId,
+        @Nonnull final ConcurrentModificationException concurrentModificationException) {
+
+        final String errorMessage = format("The payment with id '%s' couldn't be processed after %s retries.",
+            paymentId, RETRIES_LIMIT);
+        logger.error(errorMessage, concurrentModificationException);
+        return new PaymentHandleResult(HttpStatusCode.ACCEPTED_202, errorMessage);
+    }
+
+    private PaymentHandleResult handleNotFoundException(@Nonnull String paymentId, @Nonnull Exception exception ) {
         final String body = format("Failed to process the commercetools Payment with id [%s], as the payment or the cart could not be found.", paymentId);
         logger.error(tenantNameKeyValue, body, exception);
         return new PaymentHandleResult(HttpStatusCode.NOT_FOUND_404, body);
--- a/service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
@@ -91,8 +91,8 @@ public class PaymentHandler {
             .getPaymentInterface();
 
         if (!payoneInterfaceName.equals(paymentInterface)) {
-            final String errorMessage = format("Wrong payment interface name: expected '%s', found '%s'",
-                payoneInterfaceName, paymentInterface);
+            final String errorMessage = format("Wrong payment interface name: expected '%s', found '%s' for the "
+                + "commercetools Payment with id '%s'.", payoneInterfaceName, paymentInterface, paymentId);
             return new PaymentHandleResult(HttpStatusCode.BAD_REQUEST_400, errorMessage);
         }
 
@@ -126,7 +126,7 @@ public class PaymentHandler {
         logger.error(tenantNameKeyValue,
             format("Failed to process the commercetools Payment with id [%s] due to an error response from the commercetools platform.", paymentId), e);
         return new PaymentHandleResult(e.getStatusCode(),
-                format("Failed to process the commercetools payment with id [%s], due to an error response from the "
+                format("Failed to process the commercetools Payment with id [%s], due to an error response from the "
                     + "commercetools platform. Try again later.", paymentId));
     }
 
--- a/service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
@@ -91,10 +91,9 @@ public class PaymentHandler {
             .getPaymentInterface();
 
         if (!payoneInterfaceName.equals(paymentInterface)) {
-            logger.warn(tenantNameKeyValue,
-                "Wrong payment interface name: expected [{}], found [{}]",
+            final String errorMessage = format("Wrong payment interface name: expected '%s', found '%s'",
                 payoneInterfaceName, paymentInterface);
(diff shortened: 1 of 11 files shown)


service/src/main/java/com/commercetools/pspadapter/payone/notification/NotificationProcessorBase.java
```diff
@@ -121,23 +120,20 @@ public final void processTransactionStatusNotification(final Notification notifi
...
-        } else {
-            // This may happen when Payone sends the notification faster then user was redirected back to the shop
-            // and the order has not been created yet.
-            logger.info("Payment [{}] is updated, but respective order is not found!", updatedPayment.getId());
+
+        final PaymentState newPaymentState = getPaymentToOrderStateMapper()
+            .mapPaymentToOrderState(updatedPayment);
+
+        if (newPaymentState == null && updatedPayment.getPaymentStatus() != null) {
+            // TODO do we need this warning?
```
heshamMassoud: @butenkor @ahalberkamp ☝️

heshamMassoud: I will create an issue out of it, so it can be addressed later.

service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
```diff
@@ -55,68 +52,91 @@ public PaymentHandler(String payoneInterfaceName, String tenantName,
...
-                }
-            }
-
-            logger.error(
-                tenantNameKeyValue,
-                "The payment [{}] couldn't be processed after {} retries", paymentId, RETRIES_LIMIT);
-            return new PaymentHandleResult(HttpStatusCode.ACCEPTED_202,
-                    format("The payment couldn't be processed after %s retries", RETRIES_LIMIT));
-
+            return attemptToProcessPayment(paymentId, 0);
```
lojzatran:
```suggestion
            return processPaymentWithRetry(paymentId, 0);
```

service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
```diff
@@ -55,68 +52,91 @@ public PaymentHandler(String payoneInterfaceName, String tenantName,
...
+        final String paymentInterface = paymentWithCartLike
+            .getPayment()
+            .getPaymentMethodInfo()
+            .getPaymentInterface();
+
+        if (!payoneInterfaceName.equals(paymentInterface)) {
+            logger.warn(tenantNameKeyValue,
+                "Wrong payment interface name: expected [{}], found [{}]",
+                payoneInterfaceName, paymentInterface);
+            return new PaymentHandleResult(HttpStatusCode.BAD_REQUEST_400);
```
lojzatran: Should you also inform the customer about the error?

service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
```diff
@@ -55,68 +52,91 @@ public PaymentHandler(String payoneInterfaceName, String tenantName,
...
+    }
+
+    private PaymentHandleResult handleConcurrentModificationException(
+        @Nonnull final String paymentId,
+        @Nonnull final ConcurrentModificationException concurrentModificationException) {
+
+        final String errorMessage = format("The payment with id '%s' couldn't be processed after %s retries.",
+            paymentId, RETRIES_LIMIT);
+        logger.error(errorMessage, concurrentModificationException);
+        return new PaymentHandleResult(HttpStatusCode.ACCEPTED_202, errorMessage);
```
lojzatran: How does it work in this case? 202 HTTP status is returned to user, so it means at some point later it will be processed? Who will process it later?

service/src/main/java/com/commercetools/pspadapter/payone/notification/NotificationProcessorBase.java
heshamMassoud: https://github.com/commercetools/commercetools-payone-integration/issues/259

heshamMassoud: It's quite nosiy.

service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
heshamMassoud: Actually I am also not sure here because it was already like that. @butenkor any idea why it used to return 202 here? When will it be acted upon later?

> The HyperText Transfer Protocol (HTTP) 202 Accepted response status code indicates that the request has been received but not yet acted upon.

service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
heshamMassoud: I removed the warning here at all and put the error body in the response: 0e934f2

service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
heshamMassoud: e43a9ac

service/src/main/java/com/commercetools/pspadapter/payone/ScheduledJob.java
```diff
@@ -42,17 +41,24 @@ public void execute(final JobExecutionContext context) throws JobExecutionExcept
...
+                        ex);
                 } catch (final ConcurrentModificationException ex) {
-                    LOG.info(LOG_PREFIX + " Could not dispatch payment with ID \"{}\": The payment is currently processed by someone else.",
-                            tenantFactory.getTenantName(), payment.getId());
-                } catch (final Throwable ex) {
-                    LOG.error(LOG_PREFIX + " Error dispatching payment with ID \"{}\"",
-                            tenantFactory.getTenantName(), payment.getId(), ex);
+                    LOG.info(
+                        createTenantKeyValue(tenantFactory.getTenantName()),
+                        format("Could not dispatch payment with id '%s': The payment is currently processed by someone else.", payment.getId()),
```
heshamMassoud: We don't need this. Why do we have it @butenkor?

heshamMassoud:
@lojzatran 
test coverage: d2bb89a

service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
```diff
@@ -55,68 +52,90 @@ public PaymentHandler(String payoneInterfaceName, String tenantName,
...
+            if (currentAttemptCount == RETRIES_LIMIT - 1) {
+                throw concurrentModificationException;
+            } else {
+                Thread.sleep(RETRY_DELAY);
+                return processPaymentOrRetry(paymentId, currentAttemptCount + 1);
+            }
+        }
+    }
+
+    private PaymentHandleResult processPayment(@Nonnull final String paymentId) {
```
ahalberkamp: Maybe add a throw declaration here.

service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
```diff
@@ -55,68 +52,90 @@ public PaymentHandler(String payoneInterfaceName, String tenantName,
...
+        final int currentAttemptCount) throws InterruptedException {
+
+        try {
+            return processPayment(paymentId);
+        } catch (final ConcurrentModificationException concurrentModificationException) {
+            if (currentAttemptCount == RETRIES_LIMIT - 1) {
+                throw concurrentModificationException;
+            } else {
+                Thread.sleep(RETRY_DELAY);
+                return processPaymentOrRetry(paymentId, currentAttemptCount + 1);
```
butenkor: The rewrite to recursion makes code more elegant but introduces other issues as in general it consumes more memory than simple iteration (worst case scenario if it loops long enough it will run into StackOverflow error) and is slower.

service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
butenkor: There are 2 build-in cron jobs which process not processed payments. Those run independently from handle request called externally. It means that after the caller got a 202 a minute later a cron could successfully process the payment.

service/src/main/java/com/commercetools/pspadapter/payone/ScheduledJob.java
```diff
@@ -42,17 +41,24 @@ public void execute(final JobExecutionContext context) throws JobExecutionExcept
...
+                    final PaymentWithCartLike paymentWithCartLike = queryExecutor
+                        .getPaymentWithCartLike(payment.getId(), CompletableFuture.completedFuture(payment));
                     paymentDispatcher.dispatchPayment(paymentWithCartLike);
                 } catch (final NoCartLikeFoundException ex) {
-                    LOG.debug(LOG_PREFIX + " Could not dispatch payment with ID \"{}\": {}",
-                            tenantFactory.getTenantName(),  payment.getId(), ex.getMessage());
+                    LOG.error(
+                        createTenantKeyValue(tenantFactory.getTenantName()),
+                        format("Could not dispatch payment with id '%s'", payment.getId()),
+                        ex);
```
butenkor: I am not sure it is useful to see a stack trace in case cart is not found. Why following
`Could not dispatch payment with id 123 as its cart could not be found.` is not enough?

service/src/main/java/com/commercetools/pspadapter/payone/ScheduledJob.java
butenkor: I guess, it can be removed. Are you going just to swallow here?

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/PayonePostServiceImpl.java
```diff
@@ -57,15 +57,9 @@ public static PayonePostServiceImpl of(final String payoneServerApiUrl) throws I
                     getNameValuePairsWithExpandedLists(baseRequest.toStringMap(false));
 
             String serverResponse = executePostRequestToString(this.serverAPIURL, mappedListParameters);
-
-            if (serverResponse.contains("status=ERROR")) {
-                LOG.error("-> Payone POST request parameters: {}",
-                        getNameValuePairsWithExpandedLists(baseRequest.toStringMap(true)).toString());
-                LOG.error("Payone POST response: {}", serverResponse);
-            }
```
butenkor: I think POST response might be useful or the idea is that one can look it up on CTP payment object?

service/src/main/java/com/commercetools/pspadapter/payone/notification/NotificationProcessorBase.java
butenkor: It seems not necessary but it also means that we do not have a complete mapping here: https://github.com/commercetools/commercetools-payone-integration/blob/5859ac7c8254aa7514f18e491b3d6fee24351941/service/src/main/java/com/commercetools/pspadapter/payone/mapping/order/DefaultPaymentToOrderStateMapper.java#L23
In other words Payone might introduce new status that we can not map yet and thus having wrong status set. We would need to look at some examples to see if this log is required or not.

service/src/main/java/com/commercetools/pspadapter/payone/notification/NotificationProcessorBase.java
```diff
@@ -121,23 +120,20 @@ public final void processTransactionStatusNotification(final Notification notifi
...
-            if (newPaymentState != null && !newPaymentState.equals(order.getPaymentState())) {
-                return getOrderService().updateOrderPaymentState(order, newPaymentState);
-            }
-        } else {
-            // This may happen when Payone sends the notification faster then user was redirected back to the shop
-            // and the order has not been created yet.
-            logger.info("Payment [{}] is updated, but respective order is not found!", updatedPayment.getId());
+
+        final PaymentState newPaymentState = getPaymentToOrderStateMapper()
+            .mapPaymentToOrderState(updatedPayment);
```
butenkor: I know it is not much happening here but previously we would not execute this line if order was `null`

service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
heshamMassoud: Do you think it makes sense to add this info to the log message?

service/src/main/java/com/commercetools/pspadapter/payone/ScheduledJob.java
heshamMassoud: the thing is the google error reporting only works (only renders a log as an error in the error reporting when it has a stack trace otherwise, it's not picked up in the error reporting errors).

https://cloud.google.com/error-reporting/docs/formatting-error-messages

service/src/main/java/com/commercetools/pspadapter/payone/ScheduledJob.java
heshamMassoud: No, it will be caught with the more generic Exception in the catch block below and it will be logged as an error with its stack trace: cbc2a81

Would this be a problem?

service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
heshamMassoud: That is true but in this case it's only for the `RETRIES_LIMIT = 20` so its not significant and cannot lead to Stackoverflow. I would choose code readability in that case. WDYT?

service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
heshamMassoud: yep, I agree. Even though we should not declare a throwing an unchecked exception. But in this case, it's a good hint for the called of the method that they should catch it.

heshamMassoud: https://github.com/commercetools/commercetools-payone-integration/pull/257/commits/1c02b6420fc08be0586c362b1e48ba950f2d338a

service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
heshamMassoud: 18485e5

PR #265, opened 2020-01-13
Don't shorten stack trace.

heshamMassoud:
don't shorten stack trace on error logs.

### DIFF
--- a/service/src/main/resources/logback.xml
+++ b/service/src/main/resources/logback.xml
@@ -31,12 +31,6 @@ As such, enabling logback status data is very highly recommended and should be c
                 <stackTrace>
                     <!-- Map default "stack_trace" to "stackTrace" -->
                     <fieldName>stackTrace</fieldName>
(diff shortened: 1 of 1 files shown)


service/src/main/resources/logback.xml
```diff
@@ -31,12 +31,6 @@ As such, enabling logback status data is very highly recommended and should be c
                 <stackTrace>
                     <!-- Map default "stack_trace" to "stackTrace" -->
                     <fieldName>stackTrace</fieldName>
-                    <throwableConverter class="net.logstash.logback.stacktrace.ShortenedThrowableConverter">
-                        <maxDepthPerThrowable>30</maxDepthPerThrowable>
-                        <maxLength>2048</maxLength>
-                        <shortenedClassNameLength>20</shortenedClassNameLength>
-                        <rootCauseFirst>true</rootCauseFirst>
```
butenkor: Maybe this can stay, wdyt?

PR #300, opened 2020-08-19
add 2 new custom type definitions for iDeal & bancontact

JudeNiroshan:
## Description

Introduced iDeal & bancontact custom type definitions.

These custom types are attached to payment objects in commercetools platform under custom attribute.

iDeal & bancontact custom fields are identical. Except one additional attribute has added for iDeal custom type definition.

- `createMultiLineStringFieldDefinition(CustomFieldKeys.BANK_COUNTRY, FieldClassifier.REQUIRED)`

### DIFF
--- a/service/src/main/java/com/commercetools/pspadapter/payone/domain/ctp/CustomTypeBuilder.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/domain/ctp/CustomTypeBuilder.java
@@ -156,8 +156,8 @@ public class CustomTypeBuilder {
         createPaymentCustomType(PAYMENT_BANK_TRANSFER_IDEAL, ImmutableList.of(
                 createSingleLineStringFieldDefinition(CustomFieldKeys.LANGUAGE_CODE_FIELD, FieldClassifier.REQUIRED),
                 createSingleLineStringFieldDefinition(CustomFieldKeys.REFERENCE_FIELD, FieldClassifier.REQUIRED),
-                createMultiLineStringFieldDefinition(CustomFieldKeys.BANK_GROUP_TYPE, FieldClassifier.REQUIRED),
-                createMultiLineStringFieldDefinition(CustomFieldKeys.BANK_COUNTRY, FieldClassifier.REQUIRED),
+                createSingleLineStringFieldDefinition(CustomFieldKeys.BANK_GROUP_TYPE, FieldClassifier.REQUIRED),
+                createSingleLineStringFieldDefinition(CustomFieldKeys.BANK_COUNTRY, FieldClassifier.REQUIRED),
                 createSingleLineStringFieldDefinition(CustomFieldKeys.REDIRECT_URL_FIELD, FieldClassifier.OPTIONAL),
                 createSingleLineStringFieldDefinition(CustomFieldKeys.SUCCESS_URL_FIELD, FieldClassifier.OPTIONAL),
                 createSingleLineStringFieldDefinition(CustomFieldKeys.ERROR_URL_FIELD, FieldClassifier.OPTIONAL),
--- a/service/src/main/java/com/commercetools/pspadapter/payone/domain/ctp/CustomTypeBuilder.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/domain/ctp/CustomTypeBuilder.java
@@ -32,6 +32,8 @@ public class CustomTypeBuilder {
     public static final String PAYMENT_CREDIT_CARD = "payment-CREDIT_CARD";
     public static final String PAYMENT_WALLET = "payment-WALLET";
     public static final String PAYMENT_BANK_TRANSFER = "payment-BANK_TRANSFER";
+    public static final String PAYMENT_BANK_TRANSFER_IDEAL = "payment-BANK_TRANSFER_IDEAL";
+    public static final String PAYMENT_BANK_TRANSFER_BANCONTACT = "payment-BANK_TRANSFER_BANCONTACT";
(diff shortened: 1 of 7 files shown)


service/src/main/java/com/commercetools/pspadapter/payone/domain/ctp/paymentmethods/PaymentMethod.java
```diff
@@ -63,6 +63,16 @@
      */
     INVOICE_KLARNA(MethodKeys.INVOICE_KLARNA);
 
+    /**
+     * @see MethodKeys#BANK_TRANSFER_IDEAL
+     */
+//    BANK_TRANSFER_IDEAL(MethodKeys.BANK_TRANSFER_IDEAL),
```
lojzatran, line 69: Why you need to comment it out?

service/src/main/java/com/commercetools/pspadapter/payone/domain/ctp/CustomTypeBuilder.java
```diff
@@ -151,6 +153,31 @@ private void createPaymentProviderAgnosticTypes() {
                 createSingleLineStringFieldDefinition(CustomFieldKeys.BIC_FIELD, FieldClassifier.OPTIONAL)
         ));
 
+        createPaymentCustomType(PAYMENT_BANK_TRANSFER_IDEAL, ImmutableList.of(
+                createSingleLineStringFieldDefinition(CustomFieldKeys.LANGUAGE_CODE_FIELD, FieldClassifier.REQUIRED),
+                createSingleLineStringFieldDefinition(CustomFieldKeys.REFERENCE_FIELD, FieldClassifier.REQUIRED),
+                createMultiLineStringFieldDefinition(CustomFieldKeys.BANK_GROUP_TYPE, FieldClassifier.REQUIRED),
```
lojzatran, line 159: Why multi line?

service/src/main/java/com/commercetools/pspadapter/payone/domain/ctp/CustomTypeBuilder.java
```diff
@@ -32,6 +32,8 @@
     public static final String PAYMENT_CREDIT_CARD = "payment-CREDIT_CARD";
     public static final String PAYMENT_WALLET = "payment-WALLET";
     public static final String PAYMENT_BANK_TRANSFER = "payment-BANK_TRANSFER";
+    public static final String PAYMENT_BANK_TRANSFER_IDEAL = "payment-BANK_TRANSFER_IDEAL";
```
lojzatran, line 35: Why bank transfer? Is it like this in the docs?

lojzatran (changes requested):

service/src/main/java/com/commercetools/pspadapter/payone/domain/ctp/CustomTypeBuilder.java
JudeNiroshan: both new payment types that we are going to implement are bank transfers. This variable value is persisted in payment object but not passed down to payone payment gateway. It's just a name I came up with and has no influence on any factors.

service/src/main/java/com/commercetools/pspadapter/payone/domain/ctp/CustomTypeBuilder.java
JudeNiroshan: good catch

service/src/main/java/com/commercetools/pspadapter/payone/domain/ctp/paymentmethods/PaymentMethod.java
JudeNiroshan: Because in our existing tests, we are loading all the payment types and filling mapping those with relevant factory classes. As we still do not have factory classes created, tests are failing in the build process. So, for now I have commented them out.

JudeNiroshan: in this line -> https://github.com/commercetools/commercetools-payone-integration/pull/300/files/f0be6df591133781904201787d303e9b27dde16e#diff-3439e6d93c20fc6f9d51971b99d0cd1eR80

service/src/main/java/com/commercetools/pspadapter/payone/domain/ctp/CustomTypeBuilder.java
JudeNiroshan: a30efe8

service/src/main/java/com/commercetools/pspadapter/payone/domain/ctp/CustomTypeBuilder.java
lojzatran: OK, but naming is important :) You could name it also CREDIT_CARD_IDEAL and it's also correct from the code perspective :) How do you know they're bank transfers? Is it in the docs as bank transfers?

JudeNiroshan: This is a quote from Specification Shop Integrations pdf file that Thomas Alf has attached to the first opening JIRA issue.

>For instance, the online bank transfer type iDeal is only available in the Netherlands and the shop integrator or merchant has to ensure that it only works with that country.

So, it stated iDeal as *onlilne bank transfer*. For bancontact it is similarly. Because both of these payment type requests require ** onlinebanktransfertype** parameter to be passed in the requests to payone.

lojzatran (approved):

PR #309, opened 2020-08-24
#307 add bancontact

lojzatran:
Fixes: #307 

I also tested it manually the process and it seems to work. I received a success notification for the payment. See `paymentId="5e90dd5a-ab06-4b53-991f-f42ffcac5453"` in a project `ct-payone-integration-ci`

### DIFF
--- a/service/src/main/java/com/commercetools/pspadapter/payone/mapping/BankTransferWithoutIbanBicRequestFactory.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/mapping/BankTransferWithoutIbanBicRequestFactory.java
@@ -50,7 +50,12 @@ public class BankTransferWithoutIbanBicRequestFactory extends PayoneRequestFacto
         mapFormPaymentWithCartLike(request, paymentWithCartLike);
 
         //Despite declared as optional in PayOne Server API documentation. the Bankcountry is required
-        request.setBankcountry(request.getCountry());
+        final String bankCountry = ctPayment.getCustom().getFieldAsString(CustomFieldKeys.BANK_COUNTRY);
+        if (bankCountry == null) {
+            request.setBankcountry(request.getCountry());
+        } else {
+            request.setBankcountry(bankCountry);
+        }
         return request;
     }
 }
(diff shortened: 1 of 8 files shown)


service/src/main/java/com/commercetools/pspadapter/payone/mapping/BankTransferWithoutIbanBicRequestFactory.java
```diff
@@ -50,7 +50,12 @@ public BankTransferRequest createRequestInternal(@Nonnull final RequestType requ
         mapFormPaymentWithCartLike(request, paymentWithCartLike);
 
         //Despite declared as optional in PayOne Server API documentation. the Bankcountry is required
-        request.setBankcountry(request.getCountry());
+        final String bankCountry = ctPayment.getCustom().getFieldAsString(CustomFieldKeys.BANK_COUNTRY);
+        if (bankCountry == null) {
+            request.setBankcountry(request.getCountry());
```
JudeNiroshan, line 55: in case of bancontact, there is a probability that we are inserting the country value something other than `BE`. This makes a room from payone integration project to create wrong requests to payone. Should we instead make the custom-type definition to be compulsory for bancontact and iDeal? so, we don't require to check validation at this point. wdyt?

JudeNiroshan (changes requested):

service/src/main/java/com/commercetools/pspadapter/payone/mapping/BankTransferWithoutIbanBicRequestFactory.java
JudeNiroshan: ohh, okay, we already marked the custom type definition as mandatory. So, this makes sense now 👍

JudeNiroshan (approved):

service/src/main/java/com/commercetools/pspadapter/payone/mapping/BankTransferWithoutIbanBicRequestFactory.java
lojzatran: Do you mean to put instead of type "String" you would put some other type for country attribute in the product type? Is there a type "country"?

ahalberkamp (approved):

PR #301, opened 2020-08-20
Add retry decorator

leungkinghin:
RetrySphereClientDecorator is needed for recoverable errors on the payment service, as well as the request to CTP.

PR #315, opened 2020-08-26
Add retry decorator v2

leungkinghin:
RetrySphereClientDecorator is needed for recoverable errors on the payment service, as well as the request to CTP.

This is a newly created PR. The previous PR can be found here
#301 

It is because other fix for functional tests has been applied on master branch in these days. Featured branch has to be forked from master branch again to pass functional tests.


butenkor (approved):
👍

PR #323, opened 2020-09-21
312: remove unnecessary iterations on successful payment processing

JudeNiroshan:
## Description

- Fixed unnecessary iterations on successful payment handling
- Reduce retry count from 20 to 5
- Add variable delay for retrying mechanism

### DIFF
--- a/service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
@@ -12,6 +12,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import java.util.ConcurrentModificationException;
+import java.util.Random;
 
 import static com.commercetools.pspadapter.tenant.TenantLoggerUtil.createTenantKeyValue;
 import static io.sphere.sdk.http.HttpStatusCode.INTERNAL_SERVER_ERROR_500;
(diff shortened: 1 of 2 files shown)


service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
```diff
@@ -73,7 +73,8 @@ public PaymentHandleResult handlePayment(@Nonnull final String paymentId) {
         } catch (final Exception e) {
             return handleException(e, paymentId);
```
ahalberkamp, line 74: Does it make sense to log the retrycounter here too

JudeNiroshan: @ahalberkamp yeah, totally. My previous git push has missed some changelogs. Please check now :)

ahalberkamp (approved):

service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
```diff
@@ -51,29 +52,30 @@ public PaymentHandler(String payoneInterfaceName, String tenantName,
      * @return the result of handling the payment
      */
     public PaymentHandleResult handlePayment(@Nonnull final String paymentId) {
-        PaymentHandleResult result = null;
+        int retryCounter = 0;
         try {
-            for (int i = 0; i < RETRIES_LIMIT; i++) {
+            for (; retryCounter < RETRIES_LIMIT; retryCounter++) {
```
ninalindl, line 57: Isn't the first semicolon unnecessary?

ninalindl (approved):

service/src/main/java/com/commercetools/pspadapter/payone/PaymentHandler.java
JudeNiroshan: it's good you point it out. Unfortunately we need this. In java for loop, we must have a statement/assignment as the first parameter to the for loop. Therefore we cannot remove that semi-colon. We can do it in GoLang 😄

PR #329, opened 2020-10-05
Add retry predicate and unit test

leungkinghin:
Add retry predicate so that retry mechanism can be triggered when HTTP exception occurs during sending request to CTP.

### DIFF
--- a/service/src/main/java/com/commercetools/util/SphereClientConfigurationUtil.java
+++ b/service/src/main/java/com/commercetools/util/SphereClientConfigurationUtil.java
@@ -14,6 +14,7 @@ import io.sphere.sdk.http.HttpException;
 
 import javax.annotation.Nonnull;
 import java.time.Duration;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Random;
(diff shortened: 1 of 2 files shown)


ahalberkamp (approved):

service/src/main/java/com/commercetools/util/SphereClientConfigurationUtil.java
```diff
@@ -57,11 +59,17 @@ protected static BlockingSphereClient decorateSphereClient(@Nonnull final Sphere
...
+
+        final RetryPredicate httpErrorMatcher =
+                RetryPredicate.ofMatchingErrors(io.sphere.sdk.http.HttpException.class);
+
         final RetryPredicate http5xxMatcher =
-                RetryPredicate.ofMatchingStatusCodes(
-                        BAD_GATEWAY_502, SERVICE_UNAVAILABLE_503, GATEWAY_TIMEOUT_504);
+                RetryPredicate.ofMatchingStatusCodes(BAD_GATEWAY_502, SERVICE_UNAVAILABLE_503, GATEWAY_TIMEOUT_504);
+
+        final Predicate<RetryContext> retryPredicate = http5xxMatcher.or(httpErrorMatcher);
```
JudeNiroshan, line 69: chaining the exceptionMatcher with our normal HTTP retry will cause to add variable delay for both matches. If we have this random Exception happening, the customer will wait 10 seconds to complete the payment. I would not add any delay if it is a HttpException. All we want to make sure is that, if there is a HttpException, we should retry our request in a different connection from the HTTP connection pool.

JudeNiroshan (changes requested):

service/src/main/java/com/commercetools/util/SphereClientConfigurationUtil.java
leungkinghin: https://github.com/commercetools/commercetools-payone-integration/pull/329/commits/d74bb7c925f1a3371f4078f61fbe4847c9dc3e3a

ahalberkamp (approved):

service/src/main/java/com/commercetools/util/SphereClientConfigurationUtil.java
```diff
@@ -57,11 +56,20 @@ protected static BlockingSphereClient decorateSphereClient(@Nonnull final Sphere
...
+                RetryPredicate.ofMatchingErrors(io.sphere.sdk.http.HttpException.class);
+
         final RetryPredicate http5xxMatcher =
-                RetryPredicate.ofMatchingStatusCodes(
-                        BAD_GATEWAY_502, SERVICE_UNAVAILABLE_503, GATEWAY_TIMEOUT_504);
-        final List<RetryRule> retryRules =
-                Collections.singletonList(RetryRule.of(http5xxMatcher, scheduledRetry));
+                RetryPredicate.ofMatchingStatusCodes(BAD_GATEWAY_502, SERVICE_UNAVAILABLE_503, GATEWAY_TIMEOUT_504);
+
+        List<RetryRule> retryRules = new ArrayList<>();
```
JudeNiroshan, line 69:
```suggestion
        final List<RetryRule> retryRules = new ArrayList<>(2);
```

JudeNiroshan (approved):
Minimalistic optimization is possible.

PR #367, opened 2021-04-13
Replace Guava types and utilities

salander85:
This contains replacement of Guava types and utilities by using Java 8 functionalities and types.
- Exception: Guavas Caching library will be removed in another branch.

### DIFF
--- a/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
@@ -5,9 +5,14 @@ import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Splitter;
+import org.apache.commons.lang3.StringUtils;
+import spark.utils.CollectionUtils;
 
 import java.io.Serializable;
+import java.util.Collections;
 import java.util.Map;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 /**
  * @author fhaertig
@@ -58,11 +63,28 @@ public class Notification implements Serializable {
 
     public static Notification fromKeyValueString(final String keyValueString, final String separatorPattern) {
 
-        Map<String, String> notificationValues = Splitter.onPattern(separatorPattern).withKeyValueSeparator("=").split(keyValueString);
+        final Map<String, String> notificationValues = Pattern.compile(separatorPattern)
+               .splitAsStream(keyValueString.trim())
+               .map(s -> s.split("="))
+               .collect(Collectors.toMap(a -> a[0], a -> a.length>1? a[1]: ""));
+
+        validateNotificationValues(notificationValues);
 
         ObjectMapper mapper = new ObjectMapper();
         mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
         return mapper.convertValue(notificationValues, Notification.class);
+    }
+
+    private static void validateNotificationValues(final Map<String, String> notificationValues) {
+        if(notificationValues == null || notificationValues.isEmpty()) {
+            throw new IllegalArgumentException("The notification string is null or empty.");
+        }
+
+        notificationValues.values()
+                          .stream()
+                          .filter(val -> !StringUtils.isBlank(val))
+                          .findFirst()
+                          .orElseThrow(IllegalArgumentException::new);
     }
 
     @Override
--- a/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
@@ -63,6 +63,12 @@ public class Notification implements Serializable {
 
     public static Notification fromKeyValueString(final String keyValueString, final String separatorPattern) {
 
+        //This code creates a map from an input string like this:
+        //                "key=123&" +
+        //                "txaction=appointed&" +
+        //                "blabla=23"
+        //while it splits the string using the separatorPattern (in this case its "\r?\n?&") into substrings
+        //which are splitted again at the equality-sign and collected then as key-value pairs into a map
         final Map<String, String> notificationValues = Pattern.compile(separatorPattern)
                .splitAsStream(keyValueString.trim())
                .map(s -> s.split("="))
--- a/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
@@ -4,12 +4,9 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.base.Splitter;
 import org.apache.commons.lang3.StringUtils;
-import spark.utils.CollectionUtils;
 
 import java.io.Serializable;
-import java.util.Collections;
 import java.util.Map;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -69,10 +66,12 @@ public class Notification implements Serializable {
         //                "blabla=23"
         //while it splits the string using the separatorPattern (in this case its "\r?\n?&") into substrings
         //which are splitted again at the equality-sign and collected then as key-value pairs into a map
+        //throws IllegalArgumentException on duplicate keys.
         final Map<String, String> notificationValues = Pattern.compile(separatorPattern)
                .splitAsStream(keyValueString.trim())
                .map(s -> s.split("="))
-               .collect(Collectors.toMap(a -> a[0], a -> a.length>1? a[1]: ""));
+               .collect(Collectors.toMap(a -> a[0], a -> a.length>1? a[1]: "",
+                   (oldKey, newKey) -> {throw new IllegalArgumentException(oldKey + " is not a valid entry.");}));
 
         validateNotificationValues(notificationValues);
 
--- a/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
@@ -89,7 +89,8 @@ public class Notification implements Serializable {
                           .stream()
                           .filter(val -> !StringUtils.isBlank(val))
                           .findFirst()
-                          .orElseThrow(IllegalArgumentException::new);
+                          .orElseThrow(() -> new IllegalArgumentException("The notification string contains elements "
+                              + "with empty value."));
     }
 
     @Override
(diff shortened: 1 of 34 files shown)


service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
```diff
@@ -58,13 +63,30 @@
 
     public static Notification fromKeyValueString(final String keyValueString, final String separatorPattern) {
 
-        Map<String, String> notificationValues = Splitter.onPattern(separatorPattern).withKeyValueSeparator("=").split(keyValueString);
+        final Map<String, String> notificationValues = Pattern.compile(separatorPattern)
+               .splitAsStream(keyValueString.trim())
+               .map(s -> s.split("="))
+               .collect(Collectors.toMap(a -> a[0], a -> a.length>1? a[1]: ""));
```
lojzatran, line 69: I don't understand this part. Can you add a comment to the code what it means?

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
```diff
@@ -58,13 +63,30 @@
...
+
+        validateNotificationValues(notificationValues);
```
lojzatran, line 71: Is this sth new?

lojzatran (changes requested):
Did you try to find all occurencies of `com.google`? I could still see some of them. But I'm not sure if it belongs to this PR or another one.

salander85:
@lojzatran this branch removes the other occurences of `com.google` https://github.com/commercetools/commercetools-payone-integration/pull/368

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
salander85: Actually, the recently used https://guava.dev/releases/18.0/api/docs/com/google/common/base/Splitter.html throws `IllegalArgumentException` if the specified sequence does not split into valid map entries, or if there are duplicate keys - Thats why i've added this validation.

service/src/main/java/com/commercetools/pspadapter/payone/util/PayoneHash.java
```diff
@@ -17,6 +18,6 @@
      */
     @SuppressWarnings("deprecation")
     public static String calculate(String in) {
-        return Hashing.md5().hashString(in, Charsets.UTF_8).toString();
+        return DigestUtils.md5Hex(in);
```
butenkor, line 21: This needs to be tested I guess.

lojzatran (approved):

service/src/main/java/com/commercetools/pspadapter/tenant/TenantFactory.java
```diff
@@ -194,13 +202,13 @@ protected NotificationDispatcher createNotificationDispatcher(TenantConfig tenan
         final NotificationProcessor defaultNotificationProcessor =
                 new DefaultNotificationProcessor(this, tenantConfig, transactionStateResolver);
 
-        final ImmutableMap.Builder<NotificationAction, NotificationProcessor> builder = ImmutableMap.builder();
-        builder.put(NotificationAction.APPOINTED, new AppointedNotificationProcessor(this, tenantConfig, transactionStateResolver));
-        builder.put(NotificationAction.CAPTURE, new CaptureNotificationProcessor(this, tenantConfig, transactionStateResolver));
-        builder.put(NotificationAction.PAID, new PaidNotificationProcessor(this, tenantConfig, transactionStateResolver));
-        builder.put(NotificationAction.UNDERPAID, new UnderpaidNotificationProcessor(this, tenantConfig, transactionStateResolver));
+        final Map<NotificationAction, NotificationProcessor> processorMap = new HashMap<>();
```
butenkor, line 205: Are we sure that in all replaced cases it is ok not to use immutable map anymore as Guavas version provided a thread safety? In case it is required the following replacement can be used:

https://www.baeldung.com/java-immutable-maps#unmodifiable-vs-immutable

service/src/test/java/com/commercetools/pspadapter/payone/notification/common/PaidNotificationProcessorTest.java
```diff
@@ -145,7 +155,7 @@ public void whenChargeTransactionIsSuccess_ChangeTransactionStateNOTAdded() thro
         Payment paymentAuthSuccess = testHelper.dummyPaymentOneChargeSuccess20Euro();
         Transaction firstTransaction = paymentAuthSuccess.getTransactions().get(0);
         notification.setSequencenumber(firstTransaction.getInteractionId());
-        ImmutableList<UpdateAction<Payment>> authSuccessUpdates = testee.createPaymentUpdates(paymentAuthSuccess, notification);
+        List<UpdateAction<Payment>> authSuccessUpdates = testee.createPaymentUpdates(paymentAuthSuccess, notification);
```
butenkor, line 158: There is a replacement in case it makes sense: `Collections.unmodifiableList(list)`

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
butenkor: Also, why >1 and not >0? Can `a` be null?

salander85: @lojzatran 01bd736

salander85: @butenkor the `s.split("=")` method will return an Array - which may be empty but never null; It could have only one entry (e.g. if substring is "key="), thats why its checking for `length > 1` and just put the value if its there.
I'm curious what happens if `a[0]` is empty string... i'll check that.

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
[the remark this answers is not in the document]
salander85: https://github.com/commercetools/commercetools-payone-integration/pull/367#discussion_r612501179

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
```diff
@@ -58,13 +63,30 @@
...
 
         ObjectMapper mapper = new ObjectMapper();
         mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
         return mapper.convertValue(notificationValues, Notification.class);
     }
 
+    private static void validateNotificationValues(final Map<String, String> notificationValues) {
+        if(notificationValues == null || notificationValues.isEmpty()) {
```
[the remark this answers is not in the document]
salander85: Thanks for the hint, will use `MapUtils.isEmpty()`

service/src/main/java/com/commercetools/pspadapter/payone/util/PayoneHash.java
salander85: There's a test already: https://github.com/commercetools/commercetools-payone-integration/blob/ed35e69adeb51a09e5fdd099c0ce2e8ceadb73b7/service/src/test/java/com/commercetools/pspadapter/payone/config/PayoneConfigTest.java#L73

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
salander85: a366697

service/src/main/java/com/commercetools/pspadapter/payone/transaction/BaseDefaultTransactionExecutor.java
```diff
@@ -88,35 +88,44 @@ protected PaymentWithCartLike execute(final PaymentWithCartLike paymentWithCartL
...
+                        ChangeTransactionInteractionId.of(sequenceNumber, transactionId)
+                    )
                 ));
 
         try {
             final Map<String, String> response = payonePostService.executePost(request);
 
             final String status = response.get(PayoneResponseFields.STATUS);
             if (ResponseStatus.REDIRECT.getStateCode().equals(status)) {
+                final Map<String, Object> fieldsMap2 = new HashMap<>();
```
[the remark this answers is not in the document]
salander85: 684da3f

service/src/main/java/com/commercetools/pspadapter/payone/transaction/BaseDefaultTransactionExecutor.java
```diff
@@ -88,35 +88,44 @@ protected PaymentWithCartLike execute(final PaymentWithCartLike paymentWithCartL
 
         final BaseRequest request = createRequest(paymentWithCartLike);
 
+        final Map<String, Object> fieldsMap1 = new HashMap<>();
```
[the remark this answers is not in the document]
salander85: 684da3f

service/src/test/java/com/commercetools/pspadapter/payone/notification/common/PaidNotificationProcessorTest.java
salander85: Not possible as the list is modified in subclasses... - actually there's no need to use unmodifiable lists.

service/src/main/java/com/commercetools/pspadapter/tenant/TenantFactory.java
salander85: I've checked twice, immutability isn't needed as the maps are used locally or never gets modified, only to read values from.

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
```diff
@@ -58,13 +60,38 @@
...
+            throw new IllegalArgumentException("The notification string is null or empty.");
+        }
+
+        notificationValues.values()
+                          .stream()
+                          .filter(val -> !StringUtils.isBlank(val))
+                          .findFirst()
+                          .orElseThrow(IllegalArgumentException::new);
```
butenkor, line 92: This exception has no custom messages explaining the issue?

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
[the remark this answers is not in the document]
butenkor: Just curious, do we need to validate it?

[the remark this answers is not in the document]
salander85: Actually yes, otherwise the Notification object will not be properly initialized. But i recommend to use a different approach to parse the requestBody into Notification object. This is pretty complex and its pretty hard to read...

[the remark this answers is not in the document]
butenkor: The question is what will happen if we throw an exception here? What response code will we return to payone? In case it is not 200 then Payone will retry with the same wrong payload multiple times later and this single notification might block other / valid notifications too. In case a notification is not valid and can not be processed on our end we should instead log an error and return OK/200 to Payone.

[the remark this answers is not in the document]
salander85: Current implementation logs the Error and responses with `400`: https://github.com/commercetools/commercetools-payone-integration/blob/53a72392a8b1e9e3817e38887a617cf84f6c9dc4/service/src/main/java/com/commercetools/pspadapter/payone/IntegrationService.java#L106
If there's a need to change please create a ticket.

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
salander85: 4633a9c

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/Notification.java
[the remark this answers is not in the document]
butenkor: Thanks, ticket created.

butenkor (approved):

PR #371, opened 2021-04-27
Fix security issue

leungkinghin:
Fix the vulnerable dependencies
https://app.snyk.io/org/butenkor/project/d83de7b9-abdd-4c84-baac-e40443956822
Some dependencies are direct dependencies, while most of them are transitive dependencies underlying inside jvm-sdk.

Based on official support, vulnerability report is more accurate in Snyk CLI
https://support.snyk.io/hc/en-us/articles/360002615737-Snyk-CLI-shows-no-vulnerabilities-but-Snyk-UI-does

Therefore, to verify the security fix, it is recommended to test the vulnerability in CLI :

1. Download and install Snyk CLI by Homebrew 
https://support.snyk.io/hc/en-us/articles/360003812538-Install-the-Snyk-CLI#UUID-876089c6-d195-a81e-4c7a-21354f788306
2. Checkout the source from branch 'fix-securit-issue''
3. In command prompt, input 'snyk monitor --all-projects'

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ def depVersions = [
         assertJ: '3.19.0',
         mockito: '3.9.0',
         caffeine: '2.9.0',
-        jackson: '2.11.4',
+        jackson: '2.12.1',
         jetty: '9.4.39.v20210325',
         netty : '4.1.63.Final',
         testng : '6.14.3',
@@ -64,11 +64,7 @@ subprojects {
         implementation "com.github.ben-manes.caffeine:caffeine:${depVersions.caffeine}"
         implementation "com.commercetools.sdk.jvm.core:commercetools-java-client:${depVersions.commercetoolsSdkJvm}"
         constraints {
-            implementation('com.fasterxml.jackson.dataformat:jackson-dataformat-cbor') {
-               version {
-                   require '2.12.1'
-               }
-            }
+            implementation "io.netty:netty-codec-http:${depVersions.netty}"
         }
         implementation "com.commercetools.sdk.jvm.core:commercetools-models:${depVersions.commercetoolsSdkJvm}"
         implementation "com.commercetools.sdk.jvm.core:commercetools-convenience:${depVersions.commercetoolsSdkJvm}"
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -56,15 +64,23 @@ subprojects {
...
         implementation "com.github.ben-manes.caffeine:caffeine:${depVersions.caffeine}"
         implementation "com.commercetools.sdk.jvm.core:commercetools-java-client:${depVersions.commercetoolsSdkJvm}"
+        constraints {
+            implementation "io.netty:netty-codec-http:${depVersions.netty}"
+            implementation "io.netty:netty-transport-native-epoll:${depVersions.netty}"
+        }
         implementation "com.commercetools.sdk.jvm.core:commercetools-models:${depVersions.commercetoolsSdkJvm}"
         implementation "com.commercetools.sdk.jvm.core:commercetools-convenience:${depVersions.commercetoolsSdkJvm}"
+        constraints {
+            implementation "commons-io:commons-io:${depVersions.commons}"
```
lojzatran, line 74: Did you try to update the library that we use? I guess these deps are transitative ones and we don't use them directly?

lojzatran (approved):
Good job!

build.gradle
leungkinghin: Updated the direct dependency versions even they are not reported
https://github.com/commercetools/commercetools-payone-integration/pull/371/commits/fe019b82356713c6b2b56fd18d45e28e152d78d0

> I guess these deps are transitative ones and we don't use them directly
So your idea is to exclude these transitive dependency rather than update them?

lojzatran: My idea was if it's possible, upgrade the deps we're using in our library, not the ones that we're not directly using.

lojzatran: If not, can you leave the issue for the SDKs so they upgrade the deps.

leungkinghin: @lojzatran Have checked in jvm-sdk already. Their latest version doesn't apply the latest version of commons-io, and so do other vulnerable dependencies.

We can create an issue in jvm-sdk side, but meanwhile we can fix it in our gradle build until they publish a newer release to resolve the vulnerability cases.

lojzatran: Yep, that's why I already approved the PR :)

leungkinghin: Corresponding issue has been created in jvm-sdk
https://github.com/commercetools/commercetools-jvm-sdk/issues/2137

PR #400, opened 2021-07-01
Setup github actions build 

praveenkumarct:
* Remove Travis build and add GitHub actions to build CI and CD.
* Documentation updated.

Hints:
- We are no more building docker images for every push or commit (discussed this with @butenkor already), But let me know if you see any concerns.


.github/workflows/cd.yml
```diff
@@ -0,0 +1,63 @@
+name: CD
+
+on:
+  push:
+    tags:
+      - '*.*.*'
```
ahmetoz, line 6: I'm seeing the latest succesfull CD sends a tag to docker with that:  `commercetools/commercetools-payone-integration:migrate-build-github-actions`

https://hub.docker.com/layers/commercetools/commercetools-payone-integration/migrate-build-github-actions/images/sha256-f96fd8eea17da18a1319a2352b0f6e834bb78b94de4f7814a1bbcb8582b5546f?context=explore

please ensure that as I would expect a tag something like `v2.17.0`

praveenkumarct: commercetools/commercetools-payone-integration:migrate-build-github-actions
This one I tested yesterday by using push and branch name.
Anyhow, I will test this once it is merge to master.

ahmetoz (approved):

PR #407, opened 2021-07-23
CTPI-542: usage tracking

ahalberkamp:
**Summary**

* Add SolutionInfo to append to User Agent header of JVM SDK client: db925c1
* Inject SyncSolutionInfo service in META-INF file to be loaded via reflection: d964c6e

Relevant Issues
https://jira.commercetools.com/browse/CTPI-542


gradle-scripts/set-library-version.gradle
```diff
@@ -0,0 +1,32 @@
+task setLibraryVersion {
+    description 'If the env var "GITHUB_TAG" is set, injects the value in the ' +
```
praveenkumarct, line 2: Are you providing "GITHUB_TAG" variable somewhere? Because I see it will be always with value `dev-version`.

ahalberkamp: the github-tag is createdwhen the lib is released,

praveenkumarct (approved):

Issue #411, opened 2021-08-05
Payment requests to Paypal fail due to empty parameters


lsc-dotsource:
We noticed, and confirmed with payone, that requests for Paypal wallet payment fail when empty parameters are passed.

This results in the following error from Payone:
`{“errormessage”:“Parameter {shipping_addressaddition} faulty or missing”,“customermessage”:“Ungültiger Liefer-Adresszusatz. Bitte überprüfen Sie die Angaben.“,”errorcode”:“1371”,“status”:“ERROR”}`

the relevant request is this:
`{request=authorization, country=, firstname=, city=, mid=45458, successurl=https://example.com/success?token=5E1ghZ84VREA&signature=AeTlqOq%2BOuWwOiGHwiBJ008a1PzGOkrQP%2B4cSeQJEew%3D, language=, integrator_name=commercetools-payone-integration, clearingtype=wlt, mode=test, reference=D00001017, errorurl=https://example.com/error, solution_name=commercetools-platform, street=, wallettype=PPE, currency=EUR, key=, email=, zip=, shipping_street=, amount=1999, shipping_company=, backurl=https://example.com/cancel, integrator_version=v2.19.0, encoding=UTF-8, api_version=3.9, lastname=, portalid=2036918, aid=45492, shipping_country=, solution_version=1}`

I believe the parameters are parsed in the MappingUtil class and assume the problem can be solved by not setting empty strings as values;
but am not certain if that would impact payment methods other than Paypal.

After talking to payone we have been told that passing empty arguments can result in unpredictable complicatons (quote in german, translation follows):
> Tatsächlich scheint PayPal hier etwas angepasst zu haben, sodass die Transaktionen mit leeren shipping-Paramtern bei PayPal abgelehnt werden. Folgendermaßen werden die Parameter von Ihnen derzeit übergeben:
> shipping_company=
> shipping_country=DE
> shipping_street=
> 
> Ich empfehle Ihnen die übergabe der o. g. Parameter für die abweichende Lieferadresse einmal zu entfernen, da die Übergabe leerer Parameter immer wieder zu unvorhersehbaren Komplikationen führen kann.

Approximate translation:
> Paypal seems to have changed something so that transaction with empty shipping parameters are refused. currently you are passing them as follows;
> shipping_company=
> shipping_country=DE
> shipping_street=
> 
> I recommend removing these parameters [for differing delivery address], as passing empty parameters can always result in unpredictable complications.

[...]

PR #413, opened 2021-09-08
Set NoShipping to 1, When shipping address  is missing.

praveenkumarct:
- Add parameter `noShipping` to WalletPaymentTransfer.
- Set value 1 if Shipping Address is missing (any of the below values).
    - PostalCode
    - shippingCountry
    - shipping_street
    - city

related issue: https://github.com/commercetools/commercetools-payone-integration/issues/411

### DIFF
--- a/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/AuthorizationRequest.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/AuthorizationRequest.java
@@ -120,6 +120,7 @@ public class AuthorizationRequest extends BaseRequest {
 
     private String backurl;
 
+    private int noShipping;
 
     protected AuthorizationRequest(final PayoneConfig config, final String requestType, final String clearingtype) {
         super(config, requestType);
@@ -435,5 +436,13 @@ public class AuthorizationRequest extends BaseRequest {
 
     public void setBackurl(final String backurl) {
         this.backurl = backurl;
+    }
+
+    public int getNoShipping() {
+        return noShipping;
+    }
+
+    public void setNoShipping(int noShipping) {
+        this.noShipping = noShipping;
     }
 }
--- a/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/AuthorizationRequest.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/AuthorizationRequest.java
@@ -120,8 +120,6 @@ public class AuthorizationRequest extends BaseRequest {
 
     private String backurl;
 
-    private int noShipping;
-
     protected AuthorizationRequest(final PayoneConfig config, final String requestType, final String clearingtype) {
         super(config, requestType);
 
@@ -436,13 +434,5 @@ public class AuthorizationRequest extends BaseRequest {
 
     public void setBackurl(final String backurl) {
         this.backurl = backurl;
-    }
-
-    public int getNoShipping() {
-        return noShipping;
-    }
-
-    public void setNoShipping(int noShipping) {
-        this.noShipping = noShipping;
     }
 }
--- a/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/wallet/WalletPreauthorizationRequest.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/wallet/WalletPreauthorizationRequest.java
@@ -13,10 +13,13 @@ public class WalletPreauthorizationRequest extends AuthorizationRequest {
 
     private String wallettype;
 
-    public WalletPreauthorizationRequest(final PayoneConfig config, final ClearingType clearingType) {
+    private int noShipping;
+
+    public WalletPreauthorizationRequest(final PayoneConfig config, final ClearingType clearingType, final int noShipping) {
         super(config, RequestType.PREAUTHORIZATION.getType(), clearingType.getPayoneCode());
 
         this.wallettype = clearingType.getSubType();
+        this.noShipping = noShipping;
     }
 
     //**************************************************************
@@ -27,4 +30,9 @@ public class WalletPreauthorizationRequest extends AuthorizationRequest {
     public String getWallettype() {
         return wallettype;
     }
+
+    public int getNoShipping() {
+        return noShipping;
+    }
+
 }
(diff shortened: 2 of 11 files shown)


service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/AuthorizationRequest.java
```diff
@@ -120,6 +120,7 @@
 
     private String backurl;
 
+    private int noShipping;
```
butenkor, line 123: This is PayPal specific and not common for all payment methods, right? Maybe put it rather into WalletAuthorizationRequest and WalletPreauthorizationRequest classes instead.

service/src/main/java/com/commercetools/pspadapter/payone/mapping/MappingUtil.java
```diff
@@ -129,22 +129,30 @@ public static void mapCustomerToRequest(@Nonnull final AuthorizationRequest requ
...
-            request.setShipping_state(shippingAddress.getState());
+        final String shippingCountry = shippingAddress.getCountry().toLocale().getCountry();
+        final String shipping_street = joinStringsIgnoringNull(asList(shippingAddress.getStreetName(),
+            shippingAddress.getStreetNumber()));
+
+        if(shippingAddress == null
+            || StringUtils.isBlank(shippingAddress.getPostalCode())
+            || StringUtils.isBlank(shippingCountry)
+            || StringUtils.isBlank(shipping_street)
+            || StringUtils.isBlank(shippingAddress.getCity())
```
butenkor, line 140: Are those 4 fields mandatory for PayPal and would result in an error from the payment provider side?

praveenkumarct: Yes, These 4 fields are mandatory, if any field is missing in this 4 then it fails, tested and confirmed using prod credentials.

service/src/main/java/com/commercetools/pspadapter/payone/mapping/MappingUtil.java
```diff
@@ -129,22 +129,30 @@ public static void mapCustomerToRequest(@Nonnull final AuthorizationRequest requ
...
+        ) {
+            request.setNoShipping(1);
```
butenkor, line 142: I wonder if it is correct to set it for all payment methods in general? Should not this be set for PayPal payments only?

praveenkumarct: https://github.com/commercetools/commercetools-payone-integration/pull/413/commits/6f1625eeeb82416971e4847793c27599ec452c7e

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/AuthorizationRequest.java
praveenkumarct: https://github.com/commercetools/commercetools-payone-integration/pull/413/commits/6f1625eeeb82416971e4847793c27599ec452c7e

service/src/main/java/com/commercetools/pspadapter/payone/mapping/MappingUtil.java
```diff
@@ -127,25 +128,41 @@ public static void mapCustomerToRequest(@Nonnull final AuthorizationRequest requ
...
+            request.setShipping_city(shippingAddress.getCity());
+            request.setShipping_country(shippingAddress.getCountry().toLocale().getCountry());
+            request.setShipping_company(joinStringsIgnoringNull(Arrays.asList(shippingAddress.getCompany(),
+                shippingAddress.getDepartment())));
+
+            if (countriesWithStateAllowed.contains(shippingAddress.getCountry())) {
+                request.setShipping_state(shippingAddress.getState());
+            }
+        } else if(!MethodKeys.WALLET_PAYPAL.equals(paymentMethod)) {
             throw new IllegalArgumentException("Missing shipping address details");
```
butenkor, line 150: Is shippingAddress required for all other payment methods?

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/wallet/WalletPreauthorizationRequest.java
```diff
@@ -13,10 +13,13 @@
 
     private String wallettype;
 
-    public WalletPreauthorizationRequest(final PayoneConfig config, final ClearingType clearingType) {
+    private int noShipping;
```
butenkor, line 16: Why do we need this variable at all? Since mappingUtil is used for all payment methods why we can not just check there if method is paypal and then based on the address data set the noShipping info.

service/src/main/java/com/commercetools/pspadapter/payone/mapping/MappingUtil.java
praveenkumarct: Good question.
Same like Paypal payments we have to check all other Payment methods documentation for understanding how to process payments for digital items.
At least this is not part of this task.

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/wallet/WalletPreauthorizationRequest.java
praveenkumarct: Sorry this comment is bit confusing for me.

In Java, we cannot set value to a variable without declaring it. 
To achieve as you said in this comment, 
- We have to declare the 'noShipping'  variable either in **Authorization** request object(which is common request object for all payment types). which is not good because this field is then available for all payment methods.
- So, as you rightly pointed out before, I have created this variable separately only for Wallet Payment method.

service/src/test/java/com/commercetools/main/BasicPaymentRequestWorkflow.java
```diff
@@ -37,10 +64,10 @@
 public class BasicPaymentRequestWorkflow {
```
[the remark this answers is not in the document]
praveenkumarct: This is an IT test, I renamed it to **BasicPaymentRequestWorkflowIT**.
And also I tried moving it to a separate package IntegrationTests, It was not easy.
I will create a task to correct the project structure and move all IT tests to IntegrationTest folder.

https://github.com/commercetools/commercetools-payone-integration/pull/413/commits/9d1d6af4cdadb371319debfb31e4c016d82dde7c

butenkor (approved):

PR #427, opened 2021-11-19
Add new field to klarna payment request.

ahalberkamp:
In this pr a new field  was added to the payone-request.
For adding this a special "field mapping" was added because the customobject field "add_paydata" has to be mapped to the request-parameter-name "add_paydata[action]"

### DIFF
--- a/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/PayonePostServiceImpl.java
+++ b/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/PayonePostServiceImpl.java
@@ -3,6 +3,7 @@ package com.commercetools.pspadapter.payone.domain.payone;
 import com.commercetools.pspadapter.payone.domain.payone.exceptions.PayoneException;
 import com.commercetools.pspadapter.payone.domain.payone.model.common.BaseRequest;
 import com.commercetools.util.PayoneHttpClientUtil;
+import jdk.nashorn.internal.ir.annotations.Immutable;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.Consts;
 import org.apache.http.HttpEntity;
@@ -28,6 +29,7 @@ import javax.annotation.Nullable;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/PayonePostServiceImpl.java
```diff
@@ -3,6 +3,7 @@
 import com.commercetools.pspadapter.payone.domain.payone.exceptions.PayoneException;
 import com.commercetools.pspadapter.payone.domain.payone.model.common.BaseRequest;
 import com.commercetools.util.PayoneHttpClientUtil;
+import jdk.nashorn.internal.ir.annotations.Immutable;
```
lojzatran, line 6: What is this import?

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/PayonePostServiceImpl.java
```diff
@@ -222,18 +224,19 @@ public static HttpResponse executeGetRequest(@Nonnull String url) throws IOExcep
     List<BasicNameValuePair> getNameValuePairsWithExpandedLists(Map<String, Object> parameters) {
         return parameters.entrySet().stream()
                 .flatMap(entry -> {
+                    String key = KEY_MAPPING.getOrDefault(entry.getKey(), entry.getKey());
```
lojzatran, line 227: I don't understand why you add this KEY_MAPPING to this method? What is the expected result you want to see returned?

ahalberkamp: His mapping should map the customfield name "paymentData" to the name of the payone request parameter "add_paydata[action}". I cannot use "add_paydata[action]" as customfieldname, because "][" are not allowed in the customfield names,

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/PayonePostServiceImpl.java
ahalberkamp: c1e964d56

lojzatran (approved):

service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/PayonePostServiceImpl.java
```diff
@@ -71,6 +71,7 @@
     private static final String ENCODING_UTF8 = "UTF-8";
 
     private String serverAPIURL;
+    private static final Map<String,String> KEY_MAPPING = Collections.singletonMap("paymentData", "add_paydata[action]");
```
butenkor, line 74: Why is it here and not in a Klarna specific class?

ahalberkamp: because it is a global mapping

ahalberkamp:
not needed

